PR #8, opened 2018-07-15
pass RelWithDebInfo to llvm build

wanderingbort:
resolves EOSIO/eosio.wasmsdk#7


wanderingbort:
Hrm, this is not a complete fix the binaries are still very large.

wanderingbort:
I used `RelWithDebInfo` originally because thats what the Ubuntu 18.04 maintainers use for packaging.  However, those are stripped before packaging which is the important difference.  Changed to `Release` and now things are looking better:
```
-rwxrwxr-x  1 root root   31M Jul 15 09:51 clang-7
```

PR #9, opened 2018-07-15
Fixes for DESTDIR

wanderingbort:
fixed a few things that did not support the `make install DESTDIR=` idiom.

resolves EOSIO/eosio.wasmsdk#6


external/wabt/CMakeLists.txt
```diff
@@ -394,7 +394,7 @@ if (NOT EMSCRIPTEN)
   # install
   #install(TARGETS ${WABT_EXECUTABLES} DESTINATION bin)
   install(TARGETS eosio-pp DESTINATION ${SDK_INSTALL_PREFIX}/bin)
-  install(FILES ${SDK_INSTALL_PREFIX}/bin/eosio-pp 
+  install(FILES ${WABT_BINARY_DIR}/bin/eosio-pp
```
wanderingbort: we may want to use the symlink macro we have in the top level clang install .txt here as well?

PR #41, opened 2018-08-09
_n literal operator

tbfleming:
`"foo"_n` is a shortcut for `name{eosio::string_to_name("foo")}`

The `N` macro has the following issues:
* `N(foo)` confuses new users; they wonder where the identifier `foo` is defined. They're shocked when they learn `foo` isn't an identifier.
* `N` returns a `uint64_t` instead of a `name`. This hurts type safety.


tbfleming:
@dskvr fyi

PR #56, opened 2018-08-28
Remove currency contract & append .gitignore

dskvr:
It appears that the `currency.hpp` file is included in `./libraries/eosiolib` erroneously.
- It is deprecated by `eosio.token`
- It is not part of the SDK

Additionally, **.gitignore**  did not ignore `.DS_Store`, used by Mac OS systems. 

_Edit_
Removed three deprecated doxygen files from `libraries/eosiolib`


dskvr:
Removed some deprecated doxygen related files.

larryk85:
@dskvr I couldn't rebase this to the develop branch cleanly and also couldn't cherry pick your commits.  So, I just created a new PR #63 with the same changes and one new removal.

PR #59, opened 2018-08-29
Update README.md

larryk85:


README.md
```diff
@@ -33,6 +33,97 @@ Finally, install the build
...
+- `fields` is a JSON array of objects with two fields (`name` and `type`), these express the fields of the class we are referencing.  For the method approach, these will be the arguments to your method.
+   - `name` is the field name and `type` is the type of said field.
+##### abi structure :: actions
+- ```actions``` is a JSON array of objects, where each object has three fields (`name`, `type` and `ricardian_contract`)
+- `name` is the name of the action, i.e. the name you would use with `push action`
+- `type` is the type of that action, if you are using the method style and the `EOSIO_ABI` macro, then this is the name of your action method. If you are using the struct approach, then this is simply the name of the struct.
+- `ricardian_contract` is a plain text contract that is cryptographically linked to each action, the full explanation of this is way beyond the scope of this readme.
+##### abi structure :: tables
+- ```tables``` is a JSON array of objects, where each object has five fields (`name`, `type`, `index_type`, `key_names`, `key_types`)
+- `name` is the name of the table that is given the multi_index typedef.
```
arhag: Could also mention that this is the name that is used in the `table` field of the `get_table_rows` RPC from the chain API plugin (and also the table name used in the `cleos get table` command).

README.md
```diff
@@ -53,13 +144,13 @@ For example:
 cmake_minimum_required(VERSION 3.5)
 project(test_example VERSION 1.0.0)
 
-if(WASM_ROOT STREQUAL "" OR NOT WASM_ROOT)
-    set(WASM_ROOT ${CMAKE_INSTALL_PREFIX})
+if(EOSIO_WASMSDK_ROOT STREQUAL "" OR NOT EOSIO_WASMSDK_ROOT)
+    set(EOSIO_WASMSDK_ROOT "/usr/local/eosio.wasmsdk")
 endif()
 list(APPEND CMAKE_MODULE_PATH ${WASM_ROOT}/lib/cmake)
```
arhag: `${WASM_ROOT}/lib/cmake` here should instead be `${EOSIO_WASMSDK_ROOT}/lib/cmake`, right?

README.md
```diff
@@ -33,6 +33,97 @@ Finally, install the build
...
+- ```actions``` is a JSON array of objects, where each object has three fields (`name`, `type` and `ricardian_contract`)
+- `name` is the name of the action, i.e. the name you would use with `push action`
+- `type` is the type of that action, if you are using the method style and the `EOSIO_ABI` macro, then this is the name of your action method. If you are using the struct approach, then this is simply the name of the struct.
+- `ricardian_contract` is a plain text contract that is cryptographically linked to each action, the full explanation of this is way beyond the scope of this readme.
+##### abi structure :: tables
+- ```tables``` is a JSON array of objects, where each object has five fields (`name`, `type`, `index_type`, `key_names`, `key_types`)
+- `name` is the name of the table that is given the multi_index typedef.
+- `type` is the struct/class type name that is being used as the table.
+- `index_type` is always "i64", this is a holdover and will be deprecated.
+- `key_names` is a JSON array of strings and is the name of the variable of primary key followed by the name any secondary tables (indexed_by name given with mulit_index typedef)
```
arhag: We should also deprecate `key_names` just like `index_type`.

README.md
```diff
@@ -33,6 +33,97 @@ Finally, install the build
...
+- `key_types` is a JSON array of strings and is the type name associated with each element in `key_names` (the first element should always be uint64, as we only support i64 primary keys).
```
arhag: We should also deprecate `key_types` just like `index_type`.

PR #63, opened 2018-08-29
removed currency from eosiolib, exchange from examples

larryk85:
It appears that the currency.hpp file is included in ./libraries/eosiolib erroneously.

It is deprecated by eosio.token
It is not part of the SDK
Additionally, .gitignore did not ignore .DS_Store, used by Mac OS systems.

Edit
Removed three deprecated doxygen files from libraries/eosiolib and examples/exchange


larryk85:
This PR is an extension/rebase from #56

PR #67, opened 2018-08-30
Feature/abigen

larryk85:


README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
 - _eosio-cpp_, _eosio-cc_, _eosio-ld_, and _eosio-pp_ are set the core set of tools that you will interact with.
     * These are the C++ compiler, C compiler, linker and postpass tools.
 - A simple CMake interface to build EOSIO smart contracts against WasmSDK
-- ABI generator (Coming Soon)
+- ABI generator
+
+# How to use eosio-abigen
+## using with eosio-cpp
+To generate an abi with ```eosio-cpp```, the only flag you need to pass to ```eosio-cpp``` is `-abigen`, this will tell the compiler to run `eosio-abigen` after compilation and linking stages.  If the output filename is specified as a '.wasm' file with the `-o` option (e.g. \<filename\>.wasm) then eosio-cpp with tell the abi generator to create the abi with the name \<filename\>.abi, if no '.wasm' suffix is used then the resulting output filename is still \<filename\>.abi
```
gleehokie: "with tell" should be "will tell"

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+$ eosio-abigen hello.cpp -output=hello.abi
+```
+
+This will generate one file:
+* The generated abi file (hello.abi)
+
+# Difference from old abi generator
+Unlike the old abi generator tool, the new tool uses C++11 or GNU style attributes to mark ```actions``` and ```tables```.
+
+Example:
```
gleehokie: I think this would be clearer.  It moves the comment at the end of the examples to the top.

_Example (four ways to declare an action for ABI generation):_

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+    EOSLIB_SERIALIZE( testa, (n) )
+};
+
+struct __attribute__((eosio_action)) testa {
+	account_name n;
+    EOSLIB_SERIALIZE( testa, (n) )
+};
+```
+
+These are the four ways to declare an action, for abi generation.
```
gleehokie: Delete

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+
+struct __attribute__((eosio_table)) testtable {
+	uint64_t owner;
+    /* all other fields */
+};
+
+typedef eosio::multi_index<N(tablename), testtable> testtable_t;
+```
+
+These are the two ways to declare a table for abi generation.
```
gleehokie: Delete

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+
+struct __attribute__((eosio_action)) testa {
+	account_name n;
+    EOSLIB_SERIALIZE( testa, (n) )
+};
+```
+
+These are the four ways to declare an action, for abi generation.
+
+Example:
```
gleehokie: Example (two ways to declare a table for ABI generation):

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+
+struct __attribute__((eosio_table)) testtable {
+	uint64_t owner;
+    /* all other fields */
+};
+
+typedef eosio::multi_index<N(tablename), testtable> testtable_t;
+```
+
+These are the two ways to declare a table for abi generation.
```
gleehokie: Are there other examples of ABI generation besides actions and tables that you can list here?

README.md
```diff
@@ -11,7 +11,77 @@ WasmSDK is a toolchain for WebAssembly (WASM).  In addition to being a general p
...
+This will generate two files:
+* The compiled binary wasm (hello.wasm)
+* The generated abi file (hello.abi)
+
+## using eosio-abigen alone
+To generate an abi with ```eosio-abigen```, only requires that you give the main '.cpp' file to compile and the output filename `-output`.
+
+Example:
+```bash
+$ eosio-abigen hello.cpp -output=hello.abi
```
gleehokie: Why -output vs. the "usual" -o?

larryk85: -o can also be passed to eosio-abigen as part of extra C++ options (after '--' ), and is ignored if packed with them.  So, I didn't want any confusion.

README.md
larryk85: Currently, no.  Until we create a standard spec for ABI and nail down how to handle secondary indices/tables and a few other things, most of the work is done under the hood to construct these concepts we just need a better way of expressing these in the ABI.

README.md
larryk85: It's clang libtooling's option parser, so the option is -output or --output, both are interchangeable.  I used the convention of clang/gcc i.e. one dash for the example text, but if you think that would be confusing we could use eosio's convention.

PR #77, opened 2018-09-09
Fixed typos

dabdevelop:


larryk85:
These are not typos.  These showcase how to mark a struct as an eosio action, this is needed if you want to write a custom dispatcher.  You can look at the proxy contract for an example of that type of usage.  Thanks for your first-time contribution though, please continue to do so.  Also, in the future set the merge target to the ```develop``` branch instead of the ```master``` branch.

PR #133, opened 2018-10-05
Feature/ignored and contract attr

larryk85:


libraries/eosiolib/ignore.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+namespace eosio {
+   /**
+    * Tells the datastream not ignore this type, but allows the abi generator to add the correct type.
+    * @brief Tells the datastream not ignore this type, but allows the abi generator to add the correct type.
```
arhag: "Tells the datastream _to_ ignore this type, ..." for this line and the one above.

I think it is worth also putting in the comments here a statement how an `ignore` currently should not precede any non-`ignore` arguments in the action function signature and also: how that constraint is enforced by the CDT compiler and how it will be relaxed in the future.

libraries/eosiolib/datastream.hpp
```diff
@@ -328,6 +369,63 @@ inline datastream<Stream>& operator>>(datastream<Stream>& ds, eosio::symbol& sym
...
+ *  Deserialize an ignored_wrapper type from a stream
+ *
+ *  @brief Deserialize an ignore_wrapper type
+ *  @param ds - The stream to read
+ *  @param ignored - The destination for deserialized value
+ *  @tparam Stream - Type of datastream buffer
+ *  @return datastream<Stream>& - Reference to the datastream
+ */
+template<typename Stream, typename T>
+inline datastream<Stream>& operator>>(datastream<Stream>& ds, ::eosio::ignore_wrapper<T>) {
```
arhag: Remove this so we can compilation errors if someone tries to read in an `ignore_wrapper`.

PR #136, opened 2018-10-08
Feature/contract datastream

larryk85:
This pr relies on PR #133.  This adds a new field and constructor to contracts to allow for ignore types to be used in a more elegant fashion.  This pr also includes various fixes to abigen for basic functionality to allow for abi generation of eosio.contracts.


libraries/eosiolib/dispatcher.hpp
```diff
@@ -114,10 +117,10 @@ namespace eosio {
  *
  * Example:
  * @code
- * EOSIO_ABI( eosio::bios, (setpriv)(setalimits)(setglimits)(setprods)(reqauth) )
+ * EOSIO_DISPATCH( eosio::bios, (setpriv)(setalimits)(setglimits)(setprods)(reqauth) )
  * @endcode
  */
-#define EOSIO_ABI( TYPE, MEMBERS ) \
+#define EOSIO_DISPATCH( TYPE, MEMBERS ) \
```
arhag: Since dispatch logic is already changing (constructor/destructor now not called if there is no matching action) and this is a breaking minor release, I would like to clean this macro up further and get rid of the onerror stuff which really shouldn't belong here since it is a notification handler and not a regular action.

Contract developers already use custom dispatchers to handle notify handlers for things like `eosio.token::transfer`, so they could easily handle `eosio::onerror` handling (which is far less likely to be used).

Later when we have proper support for action _and_ notify handlers in CDT code generator, we will deprecate this EOSIO_DISPATCH macro anyway.

libraries/eosiolib/contract.hpp
```diff
@@ -35,7 +35,7 @@ class contract {
        * @param n - The name of this contract
        * @param ds - The datastream used 
        */
-      contract( name n, datastream<const char*> ds ):_self(n), _ds(ds) {}
+      contract( name n, name code, datastream<const char*> ds ):_self(n), _ds(ds) {}
```
arhag: Missing initialization: `_code(code)`. Similar change as above regarding doxygen comment and renaming `n`.

libraries/eosiolib/contract.hpp
```diff
@@ -26,7 +26,7 @@ class contract {
        * @brief Construct a new contract object.
        * @param n - The name of this contract
        */
-      contract( name n ):_self(n) {}
+      contract( name n, name code ):_self(n), _code(code) {}
```
arhag: Add code to doxygen comment (see comment below regarding how to describe `code`). Also I would like to rename `n` to something more meaningful like `receiver`.

libraries/eosiolib/contract.hpp
```diff
@@ -61,6 +67,13 @@ class contract {
        */
       name _self;
 
+      /**
+       * The code name of this contract
```
arhag: I would instead describe this as something like the following:

`The code name of the action this contract is processing.`

libraries/eosiolib/dispatcher.hpp
```diff
@@ -124,13 +124,9 @@ namespace eosio {
 extern "C" { \
    void apply( uint64_t receiver, uint64_t code, uint64_t action ) { \
       eosio::name self(receiver); \
-      if( action == eosio::name("onerror").value) { \
-         /* onerror is only valid if it is for the "eosio" code account and authorized by "eosio"'s "active permission */ \
-         eosio_assert(code == eosio::name("eosio").value, "onerror action's are only valid from the \"eosio\" system account"); \
-      } \
-      if( code == self.value || action == eosio::name("onerror").value ) { \
+      if( code == self.value ) { \
```
arhag: For consistency we might as well remove the line above and change this line to `if( code == receiver ) {`.

Then in `EOSIO_DISPATCH_HELPER` the call to `execute_action` would be:
`eosio::execute_action( eosio::name(receiver), eosio::name(code), &OP::elem );`

PR #162, opened 2018-10-12
Release v1.3.2

larryk85:


larryk85:
Fixes for `eosio-abigen` table generation regression. #161 
Adding missing `const` declaration qualifiers #164

PR #200, opened 2018-10-25
support std::set<eosio::permission_level>

UMU618:
If no operator <, `std::set<eosio::permission_level>` can't compile. It's used in `<eosiolib/permission.hpp>`.


wuyahuang:
@larryk85 Please check this out.

PR #206, opened 2018-10-26
Add Dockerfile for generating docker image.

ericiles:
This Dockerfile create an image based on the latest release by default. The release can be overridden by specifying ```--build-arg release=<tag>``` when building the image.


larryk85:
I was able to build a container and build contracts within Docker, so I think it's good.

PR #209, opened 2018-10-27
change build.sh 1 gigabytes per core

fingera:
I saw that the compiler uses up to 1G.
I have 7G free memory, but only use one core.
This may improve compile performence?


larryk85:
Will not merge, binary builds relieve some of this issue, you can also use make -j <whatever job count you want> from the build directory.

PR #220, opened 2018-10-31
Improve `ricardeos.py` script to fit with actual documentation about ricardian contracts

maoueh:
The documentation [1] states that ricardian contact should be in the form:

```
// Ricardian contract of the overall contract
<contract_name>_rc.md

// Ricardian contract of each action
<contract_name>.<action_name>_rc.md
```

The `scripts/ricardeos.py` was not handling this correctly anymore. Updated it to read the actions contract files using the right pattern.

Also change the way ricardian clauses are retrieved. Instead of a single clause file, we "parse" the actual ricardian contract file looking for each section of the markdown file (we only accept "H3" section (`###`)) gathering section's line along the way.

I get that this last one might be questionable. I felt however that it's quite better for the developer to specify a single file.

* [1] - https://developers.eos.io/eosio-cpp/docs/hello-world#section-hello-world-ricardian-contract

### DIFF
--- a/scripts/ricardeos/ricardeos.py
+++ b/scripts/ricardeos/ricardeos.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
+import io
 import json
+import re
 import sys
 import os.path
 import fnmatch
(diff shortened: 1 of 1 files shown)


maoueh:
I just noted https://github.com/EOS-Mainnet/governance/tree/master/eosio.system uses the "old" scheme.

So maybe it's the documentation that should be updated and not this script?

larryk85:
This tool is now deprecated and gone, use `eosio-cpp` and it's ricardian importing mechanisms.

PR #227, opened 2018-11-05
Dockerfile: added make & cmake to installed packages

maoueh:
Those two packages are required when using the "standard" scripts to compile Smart Contract since they use CMake to configure the build and the make to run it.

Test it with:

```
cd examples/hello
mkdir build
docker run --rm -it -w /app/build -v `pwd`:/app <imageId> cmake ..
docker run --rm -it -w /app/build -v `pwd`:/app <imageId> make
```

~**Note** There is no `.wasm` nor `.abi` file generated however, normal?~ Oupsss, forgot I had to run `make` at the end, code 18, sorry :P

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 ARG release=latest
 
 # Install required packages
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install openssl ca-certificates curl wget && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install make cmake openssl ca-certificates curl wget && rm -rf /var/lib/apt/lists/*
 
 # Install CDT from deb package
 ADD install_deb.sh /


docker/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ubuntu:18.04
 ARG release=latest
 
 # Install required packages
-RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install openssl ca-certificates curl wget && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install make cmake openssl ca-certificates curl wget && rm -rf /var/lib/apt/lists/*
```
maoueh:
```suggestion
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get -y install make cmake g++ openssl ca-certificates curl wget && rm -rf /var/lib/apt/lists/*
```

So currently, there is a small problem that prevents the current PR to be merged as-is. Indeed, if there is no C++ compiler installed within the Docker image, CMake will complains with the following error:

```
-- Detecting C compile features - done
CMake Error at CMakeLists.txt:2 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
```

It seems CMake still forces the detection of a working C++ compiler. I tried customizing the C++ compiler to use the bundled cross-compiling one (the C++ to WASM one). It can be done simply be moving the `find_package(eosio.cdt)` line at the top of the CMakeFile.

One could think it would solve the problem, however CMake now tries to perform the C++ compiler checks which seems to consist in compiling a sample program. But it uses a standard program which the cross-compiling compiler chokes on.

I tried setting the CMake platform to generic via `set(CMAKE_SYSTEM_NAME "Generic")` but that did not help. I tried to find how to disable or change the CMake tests, but did not find anything.

So if somehow with more knowledge of CMake could shim in, that would be much appreciated.

Right now, the simple "workaround" is to add `g++` to the list of installed package along `cmake` and `make`.  If you think it's the best way to go to start, I will update my PR to add the package.

maoueh:
I tested the changes to CMake config done in PR #233 (specifically this one https://github.com/EOSIO/eosio.cdt/pull/233/files#diff-dbd8fb33f1444780d8d3c74b7e552416).

The changes make it possible to not have a C or C++ compiler installed on the machine and still be able to configure correctly CMake **only** if `` comes before the `project` clause:

```
cmake_minimum_required(VERSION 3.5)

find_package(eosio.cdt)
project(hello_example VERSION 1.0.0)

add_contract(hello hello hello.cpp)
``` 

If I use the standard current version of the `CMakeList.txt`, the compiler as configured by `eosio.cdt` cmake package are not picked up fast enough and compiler checks fail.


```
cmake_minimum_required(VERSION 3.5)
project(hello_example VERSION 1.0.0)

find_package(eosio.cdt)

add_contract(hello hello hello.cpp)
```

jeffreyssmith2nd:
The only purpose of the Dockerfile is to install the latest binary distribution. That's why it only installs a minimal amount of dependencies, and none of the developer dependencies.

PR #243, opened 2018-11-07
rename digest to checksum; add eosio::signature

arhag:
This PR makes changes to the additions in PR #236.

Specifically, this PR includes the following changes:

- To reduce confusion, it renames the type aliases to `eosio::fixed_bytes` introduced in PR #236 as follows:
   + `eosio::digest160` is now `eosio::checksum160`
   + `eosio::digest256` is now `eosio::checksum256`
   + `eosio::digest512` is now `eosio::checksum512`

- Adds a C++ type `eosio::signature` (with serialization/deserialization support) to be used instead of the C struct `capi_signature`.
- Modifies the `eosio::recover_key` and `eosio::assert_recover_key` C++ functions introduced in PR #236 which simply wrap the functionality of the corresponding C intrinsics so that they now take the new `eosio::signature` type as an argument.

### DIFF
--- a/libraries/eosiolib/CMakeLists.txt
+++ b/libraries/eosiolib/CMakeLists.txt
@@ -3,6 +3,7 @@ file(GLOB HEADERS "*.hpp"
 
 add_library(eosio
             eosiolib.cpp
+            crypto.cpp
             ${HEADERS})
 
 target_include_directories(eosio PUBLIC
--- a/libraries/eosiolib/crypto.cpp
+++ b/libraries/eosiolib/crypto.cpp
@@ -1,0 +1,91 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
+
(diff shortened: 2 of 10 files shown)


jordigoyanes:
Now I can't hash with sha256 using fixed_bytes, it gives me error: 
```
mines.cpp:18:10: error: no matching function for call to 'sha256'
sha256( (char *)&serverseed, sizeof(serverseed), &random_num_hash );
^~~~~~
/usr/opt/eosio.cdt/1.4.1/bin/../include/eosiolib/crypto.h:137:6: note: candidate function not viable: no known
conversion from 'eosio::checksum256 *' (aka 'fixed_bytes<32> *') to 'capi_checksum256 *' for 3rd argument

void sha256( const char* data, uint32_t length, capi_checksum256* hash );
```
The sha256 is expecting a capi_checksum256(deprecated) instead of eosio::checksum256

PR #304, opened 2018-12-02
fix: additional blank line is needed between markdown and html.

lychees:


jeffreyssmith2nd:
Thanks for the contribution, however this was fixed at some point when the examples were refactored.

PR #317, opened 2018-12-11
Feature/cdt tester

larryk85:
This new functionality allows for contract code to be compiled natively.  And a new tester framework to utilize this functionality.

- eosio-cpp / eosio-ld
   - new flag `fnative` added to allow for native compiles and linking
- libnative/cdt tester
   - libc, libcxx, eosiolib are now built for native and wasm versions
   - a new library with a minimal crt and functionality to set and get intrinsics at runtime.
      - eosio_assert*, memcpy/cmp/move, softfloat and builtins are pre-implemented and fixed.
      - print* intrinsics are pre-implemented and allowable to be reset at runtime.
    - a new tester library with REQUIRE_ASSERT, CHECK_ASSERT, REQUIRE_EQUAL, CHECK_EQUAL macros/ functions defined to allow for "catching" assertions and comparing the assert message, and to pass/fail accordingly and to pass/fail based on some predicate.  The REQUIRE vs. CHECK versions are much the same as BOOST test framework, as in the REQUIRE variant will kill the test immediately and CHECK will fail the test as a whole but allow for sub checks to evaluate.


larryk85:
Superceded by PR #318

PR #320, opened 2018-12-12
compatiable with pure c.

czom:


czom:
compatiable with pure c.

larryk85:
Will not merge.  The newest direction of CDT is to segregate the C API and C++ API, so the new model will work with pure C projects.

PR #333, opened 2018-12-24
Replace references of uint256_t with checksum256

### LINKED ISSUE #332: Documentation wrongly states uint256_t as an option for secondary indices
Needs to be updated

Related #183

dskvr:
- `uint256_t` -> `checksum256` 

Resolve #332

### DIFF
--- a/libraries/eosiolib/multi_index.hpp
+++ b/libraries/eosiolib/multi_index.hpp
@@ -209,8 +209,6 @@ struct indexed_by {
  *  - uint128_t
  *  - double
  *  - long double
- *
- *  `uint256_t` is not support for use as a secondary index, instead use the following
  *  - eosio::checksum256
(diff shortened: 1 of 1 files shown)


libraries/eosiolib/multi_index.hpp
```diff
@@ -207,10 +207,12 @@ struct indexed_by {
  *  EOSIO Multi-Index table also supports up to 16 secondary indices. The type of the secondary indices could be any of:
  *  - uint64_t
  *  - uint128_t
- *  - uint256_t
  *  - double
  *  - long double
  *
+ *  `uint256_t` is not support for use as a secondary index, instead use the following
```
tbfleming: Might be better not to mention `uint256_t`, since that type isn't defined anywhere

PR #334, opened 2018-12-25
Add missing boost/throw_exception.hpp

conr2d:
Added `boost/throw_exception.hpp` to make `boost/utility/string_ref.hpp` and `boost/utility/string_view.hpp` available.


larryk85:
Will not merge.  This is going against the direction we are currently going in CDT, which is moving towards complete removal of Boost from CDT.

PR #358, opened 2019-01-14
added eosio::wasm_entry attribute for user definable entry points

larryk85:


kesar:
as a suggestion, it would be great if we link PR with issue, so we can follow the reason behind the PR

PR #383, opened 2019-01-28
Help Doxygen interpret otherwise valid C++ syntax

### LINKED ISSUE #382: Doxygen gets confused in eosiolib/fixed_bytes.cpp
Conditional inside template definition for `fixed_bytes` needs to be replaced with `std::less` so Doxygen doesn't document the `for` loop creating inaccurate albeit humorous reference docs.

dskvr:
## Fix
Problematic Code:
```
         template<typename Word, size_t NumWords,
                  typename Enable = typename std::enable_if<std::is_integral<Word>::value &&
                                                             !std::is_same<Word, bool>::value &&
                                                             sizeof(Word) < sizeof(word_t)>::type >
```
The Fix: 
```
       template<typename Word, size_t NumWords,
                  typename Enable = typename std::enable_if<std::is_integral<Word>::value &&
                                                             !std::is_same<Word, bool>::value &&
                                                             std::less( sizeof(Word), sizeof(word_t))>::type >
```
Problem occurred in two places.

## Causes
Doxygen did not interpret the `<` operator in the `sizeof(Word) < sizeof(word_t)>::type` as less than.

## Observations
This only seems to affect `<`, until issue is addressed in Doxygen, should consider a temporary convention. 

## Conclusion
Doxygen now works as expected for `fixed_bytes.hpp`

Resolves #382

### DIFF
--- a/libraries/eosiolib/fixed_bytes.hpp
+++ b/libraries/eosiolib/fixed_bytes.hpp
@@ -134,7 +134,7 @@ namespace eosio {
          template<typename Word, size_t NumWords,
                   typename Enable = typename std::enable_if<std::is_integral<Word>::value &&
                                                              !std::is_same<Word, bool>::value &&
-                                                             sizeof(Word) < sizeof(word_t)>::type >
+                                                             std::less( sizeof(Word), sizeof(word_t))>::type >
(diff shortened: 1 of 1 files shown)


dskvr:
Works 👍

PR #396, opened 2019-01-30
Regorganize API doc grouping 

dskvr:
Previously, API was grouped by C, C++ and Types. New grouping is as follows

- Core
- Contracts
- Types


dskvr:
@larryk85 Please add description to `types` group in `libraries/eosiolib/eosio.hpp`

Issue #400, opened 2019-02-01
insert and extract signed_int from datastream


snstrand:
If a signed_int with large enough absolute value is inserted into a datastream the extracted value will be off by 1. The error occurs when the signed_int is >= 2^30 or < -2^30.

PR #402, opened 2019-02-01
fix issue #400 extract signed_int from datastream

snstrand:
Resolves issue #400. Added a unit test.


libraries/eosiolib/varint.hpp
```diff
@@ -427,14 +426,12 @@ struct signed_int {
...
-      do {
+      do {                                                      //read 7 bit chunks
          ds.get(b);
          v |= uint32_t(uint8_t(b) & 0x7f) << by;
          by += 7;
       } while( uint8_t(b) & 0x80 );
-      vi.value = ((v>>1) ^ (v>>31)) + (v&0x01);
-      vi.value = v&0x01 ? vi.value : -vi.value;
-      vi.value = -vi.value;
+      vi.value= (v>>1) ^ -(v&1);                                //reverse zigzag encoding
```
larryk85: This seems a little obtuse to me.  The expression -(v&1) is not immediately apparent that it is not undefined behavior and takes some staring at to see that it is fine and what it is doing.  I think it might be better to use 1ull + ~(v&1), both should compile to the same instructions (so no loss of performance, but a little more obvious in my opinion).

PR #407, opened 2019-02-06
support plus, minus operators

changmyeong:
I want to use these operators(operator+, operator-) in multi_index.

### DIFF
--- a/libraries/eosiolib/multi_index.hpp
+++ b/libraries/eosiolib/multi_index.hpp
@@ -339,6 +339,16 @@ class multi_index
                   const T& operator*()const { return *static_cast<const T*>(_item); }
                   const T* operator->()const { return static_cast<const T*>(_item); }
 
+                  const_iterator operator+(int n) {
+                     const_iterator result(*this);
+                     return result + n;
+                  }
+
+                  const_iterator operator-(int n) {
+                     const_iterator result(*this);
+                     return result - n;
+                  }
+
                   const_iterator operator++(int){
                      const_iterator result(*this);
                      ++(*this);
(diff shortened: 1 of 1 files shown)


arhag:
We are trying to follow the C++ convention of not adding operations to the iterators that would give the developer an impression of a low cost computation when in fact its implementation requires a large amount of computation. 

To provide some context, look at the iterators for `std::vector` vs `std::list`. An [`std::vector`](https://en.cppreference.com/w/cpp/container/vector) uses a [`RandomAccessIterator`](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) which does implement the plus and minus operator. On the other hand, an [`std::list`](https://en.cppreference.com/w/cpp/container/list) uses only a [`BidirectionalIterator`](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator) which does not implement the plus and minus operators, just the increment and decrement operators.

This should hopefully make sense considering that a `vector` consists of its fixed-size elements located contiguously in memory which makes jumping from one element in the vector to any other a constant time operation (since it enables random access to any of its elements by just constructing the appropriate address). But a `list` is implemented by holding pointers to the previous and next nodes within each node of the `list`. That means the only way to, for example, add some number `N` to an iterator pointing to some element of a `list` is to call the function that advances to the next element (the advancing operation is `O(1)`) `N` number of times which  makes the overall operation `O(N)`.

The `eosio::multi_index` class is implemented in a way that makes it analogous to `std::list` in terms of traversing through the sequence of elements in any given index. (In reality it is a little worse in time complexity because the native EOSIO side must map the WebAssembly-side handle integer value to a particular object managed by `boost::multi_index`, which is not a `O(1)` operation but rather an `O(logC)` operation where `C` is the number of cached items in the iterator cache within the execution context of the action.) Implementing `const_iterator operator+(int n)` would requiring calling `const_iterator& operator++()` repeatedly `n` times.

We do not want to encourage contract developers to use such an addition operation by adding it into CDT because its existence may mislead contract developers into false belief that it is just as cheap as adding an integer to an iterator of `std::vector`. If a contract really needs such functionality, they can simply use [`std::advance`](https://en.cppreference.com/w/cpp/iterator/advance).

PR #414, opened 2019-02-11
install symlinks to /usr/local/...

larryk85:
Revert some of the changes to install symlinks to system path (`/usr/local/bin` and `/usr/local/lib/cmake/eosio.cdt`).

PR #415, opened 2019-02-11
V2.1

NorseGaud:
1. Better uninstall script
2. llvm/bin removal from $PATH
3. Using /usr/local/bin and .../lib for symlinks + Wrapped symlink creation to avoid PATH modifications
4. sudo required for symlinks


NorseGaud:
Merged in changes from https://github.com/EOSIO/eosio.cdt/pull/414

PR #456, opened 2019-03-17
Add typedefs for capi struct types

conr2d:
CAPI struct types need typedef, or they will require cumbersome `struct` keyword when being used every time.


larryk85:
This is not needed as of v1.6.x.

PR #461, opened 2019-03-19
Exclude C source files from abi generation

conr2d:
When building main contract written in "c++" with static library written in "c", the flags for compiling C files contains `-abigen` and related options, so errors like `error: invalid argument '-std=c++17' not allowed with 'C'` is emitted. This patch appends abigen options only when the type of source file is c++.


larryk85:
This is no longer needed as of v1.6.0, the stable release will be today.

PR #522, opened 2019-05-23
CHECK_PRINT bug fix

stefanzarembinski:
A bug in the CHECK_PRINT macro of the tester API: it returns error conditions always.

### First change
In files:
libraries\native\native\eosio\tester.hpp, line 77
and
libraries\native\tester.hpp, line 79

The condition std_out.index == N-1 is always false what causes misleading error messages like this --
```
error : wrong print message9 {hello_test.cpp:hello_test:52}
```
-- even if the print mesage is OK.
```
 inline bool expect_print(bool check, const std::string& li, const char (&expected)[N], F&& func, Args... args) {
   return expect_print(check, li, 
         [&](const std::string& s) { 
-           return std_out.index == N-1 &&
+           return std_out.index == N &&
            memcmp(expected, s.c_str(), N-1) == 0; }, func, args...);
```

### Second change
In files:
libraries\native\native\eosio\tester.hpp, line 68
and
libraries\native\tester.hpp, line 70

The digit 9 in the message text: it is obscure and thus misleading.
```
   if (!check)
      eosio::check(passed, std::string("error : wrong print message {"+li+"}").c_str());
   if (!passed)
-     eosio::print("error : wrong print message9 {"+li+"}\n");
+     eosio::print("error : wrong print message {"+li+"}\n");
   silence_output(disable_out);
```

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,5 @@
 
 # Build
 build/*
+*.bin
+*.ipch
--- a/libraries/native/native/eosio/tester.hpp
+++ b/libraries/native/native/eosio/tester.hpp
@@ -65,7 +65,7 @@ inline bool expect_print(bool check, const std::string& li, Pred&& pred, F&& fun
    if (!check)
       eosio::check(passed, std::string("error : wrong print message {"+li+"}").c_str());
    if (!passed)
-      eosio::print("error : wrong print message9 {"+li+"}\n");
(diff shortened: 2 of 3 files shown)


stefanzarembinski:
My mistake.
Stefan Zarembinski

PR #552, opened 2019-07-12
Fix to invalid symbol print

conr2d:
## Change Description
`symbol::print` needs to use `eosio::printl` instead of `eosio::print`. Because `eosio::print(buffer, (end-buffer))` is translated into `print(buffer); print(end-buffer)`, so redundant string is printed. 

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


PhillipHamnett:
For cross reference, here is a related question from the Stack Exchange: https://eosio.stackexchange.com/questions/4655/symbol-prints-wrong-value

jeffreyssmith2nd:

conr2d:
This change is applied by f95955990dbcc0187962c4b0d38979e778dc49b5.

PR #553, opened 2019-07-12
Add check maximum precision in symbol validation

conr2d:
## Change Description
Unlike `eosio::chain::symbol`, `eosio::symbol` in contract header doesn't check max precision in validation check. This PR adds max precision check to `eosio::symbol::is_valid` to keep consistency.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/core/eosio/symbol.hpp
+++ b/libraries/eosiolib/core/eosio/symbol.hpp
@@ -243,6 +243,11 @@ namespace eosio {
    class symbol {
    public:
       /**
+       * Maximum number of decimal places used for the symbol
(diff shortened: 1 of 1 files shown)


jeffreyssmith2nd:

arhag:
I'm thinking it might be better to leave CDT's version of asset alone (to not break any contracts that may be using large precision for whatever reason) and just relax the restrictions of the asset in EOSIO.

conr2d:
According to @arhag's comment, I close this. If necessary, please reopen.

PR #564, opened 2019-07-19
Travis Trigger step

NorseGaud:


NorseGaud:
I cancelled the builds to avoid wasting resources. I tested the trigger works which is all that's needed

PR #569, opened 2019-07-26
Fix get_active_producers

me-vadim:
get_active_producers from wasm_interface treats the second argument as size of buffer in bytes, but get_active_producers from eosio.cdt puts active_prods.size() as a second argument.

This leads to corrupted active_prods where only 21 bytes of 8*21 are filled.


jeffreyssmith2nd:

PR #577, opened 2019-08-05
Set default output filename

conr2d:
## Change Description
Fix to default output filename. (when not specified output filename by `-o`)

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


johndebord:
@conr2d Please see my comment in [#578](**url**).

conr2d:
This also fixes #582.  
Object file is added to ldopts twice by compiler_options.hpp and eosio-cpp.cpp, and it causes duplicate symbol error.

PR #603, opened 2019-08-25
Fix stack overflow caused by recursive add_type 

### LINKED ISSUE #527: Segmentation fault (core dumped)
Hello,

we found an issue (cdt 1.5.0-1.6.1)  on using construction which include array of self objects, like `std::vector<testtab>` :

```
TABLE testtab {
  uint64_t    id;
  std::vector<testtab> container; 

  auto primary_key() const {  return id; }
};
```
This construction is used in simpleassets.io (NFT standart) for attaching NFTs and FTs into others NFTs (https://github.com/CryptoLions/SimpleAssets/blob/master/include/SimpleAssets.hpp#L529)

Using CDT 1.5.0 - wasm was created but abi was failed. For deploying wasused manually modified abi file (https://github.com/CryptoLions/SimpleAssets/blob/master/SimpleAssets.abi#L539)  and result was OK (img below). 

image

With cdt 1.6.1. building is failed totaly and no wasm or abi is created.

Test project with construction:
[test.tar.gz](https://github.com/EOSIO/eosio.cdt/files/3292567/test.tar.gz)

Thanks,
Best Regards,
Bohdan

conr2d:
## Change Description

Fix #527. 

`add_type` calls itself recursively to add all types it refers to. If user-defined types refer to each other or themselves, it causes stack overflow by cycle.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/abigen.hpp
+++ b/tools/include/eosio/abigen.hpp
@@ -249,6 +249,9 @@ namespace eosio { namespace cdt {
       }
 
       void add_type( const clang::QualType& t ) {
+         if (evaluated.count(t.getTypePtr()))
+            return;
+         evaluated.insert(t.getTypePtr());
          auto type = get_ignored_type(t);
          if (!is_builtin_type(translate_type(type))) {
             if (is_aliasing(type))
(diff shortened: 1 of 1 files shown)


jeffreyssmith2nd:

PR #604, opened 2019-08-25
 Fix build fail when struct inherits from std types 

### LINKED ISSUE #541: inheriting std::variant as table data segfaults
using v1.6.1 built locally on mac

this code appears to cause a segfault in `eosio-cpp`

goodbye.cpp:
```c++
#include <eosio/eosio.hpp>
#include <eosio/name.hpp>
#include <variant>

using namespace eosio;

class [[eosio::contract]] goodbye : public eosio::contract {
   public:
      goodbye(name receiver, name code, datastream<const char*> ds)
      :eosio::contract(receiver, code, ds)
      ,_foo(receiver, receiver.value)
      {}

      struct foo_v1 {
         uint64_t primary;
      };

      struct [[eosio::table]] foo : public std::variant<foo_v1> {
         uint64_t primary_key()const {
            return std::visit([](const auto& f){return f.primary; }, *this);
         }
      };

      using foo_table = eosio::multi_index<"foos"_n, foo>;

      foo_table _foo;

      [[eosio::action]]
      void hi(name nm) {
      }
};

```

conr2d:
## Change Description

Fix #541.

`add_struct` enforces parsing its base as struct, and it causes segfault when inheriting from the std types. This patch mitigates this issue by allowing add_struct to call add_type for its base, but generated ABI is still not usable, because chain::abi_serializer requires struct base to be struct type. There is a workaround fix making it work by trick, but ultimately it needs consensus upgrade.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/abigen.hpp
+++ b/tools/include/eosio/abigen.hpp
@@ -159,7 +159,7 @@ namespace eosio { namespace cdt {
          abi_struct ret;
          if ( decl->getNumBases() == 1 ) {
             ret.base = get_type(decl->bases_begin()->getType());
-            add_struct(decl->bases_begin()->getType().getTypePtr()->getAsCXXRecordDecl());
+            add_type(decl->bases_begin()->getType());
          }
          std::string sub_name = "";
          for ( auto field : decl->fields() ) {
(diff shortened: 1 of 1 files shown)


conr2d:
This change is applied by merging #608.

PR #605, opened 2019-08-25
Fix segfault when using nested typedef

### LINKED ISSUE #600: Build fails when using nested typedef
test.cpp:
``` c++
#include <eosio/eosio.hpp>
using namespace eosio;

namespace foo {
// Both `using` and `typedef` cause segfault.
//using str = std::string;
typedef std::string str;
}

class [[eosio::contract]] test : public contract {
public:
   using contract::contract;
   [[eosio::action]]
   void hi(foo::str s) {
      print(s);
   }
};
```

console:
```
$ eosio-cpp -o test.wasm test.cpp
Warning, empty ricardian clause file
Warning, empty ricardian clause file
Segmentation fault (core dumped)
```

conr2d:
## Change Description

Fix #600.

When typedef is nested in other namespace, `clang::ElaboratedType` is passed to `get_type_alias` instead of `clang::TypedefType`. When dynamic cast to clang::TypedefType fails, try casting to
ElaboratedType once mroe to avoid build fail by segfault.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/gen.hpp
+++ b/tools/include/eosio/gen.hpp
@@ -569,12 +569,16 @@ struct generation_utils {
    inline std::string get_type_alias_string( const clang::QualType& t ) {
       if (auto dt = llvm::dyn_cast<clang::TypedefType>(t.getTypePtr()))
          return get_type(dt->desugar());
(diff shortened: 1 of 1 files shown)


jeffreyssmith2nd:

PR #606, opened 2019-08-25
Fix segfault during ABI generation for tuple

### LINKED ISSUE #558: Segfault when using a class from the std namespace.
Issues #336 and #521 highlight a problem when `using` a class from the `std` namespace. Currently, both `tuple` and `variant` have been proven to segfault.

Example code:
```cpp
#include <eosio/eosio.hpp>
#include <eosio/print.hpp>
#include <tuple>

using std::tuple;
using namespace eosio;

class[[eosio::contract("hello")]] hello : public contract
{
public:
   using contract::contract;

   [[eosio::action]] void hi(name user) {
      require_auth(user);
      print("Hello, ", user);
   }

   struct [[eosio::table]] greeting {
      uint64_t id;
      tuple<int, int> t;
      uint64_t primary_key() const { return id; }
   };
   typedef multi_index<"greeting"_n, greeting> greeting_index;
};
```

conr2d:
## Change Description

Fix #558. 

When passed parameter of clang::QualType is not clang::ElaboratedType, segfault occurs during ABI generation.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/abigen.hpp
+++ b/tools/include/eosio/abigen.hpp
@@ -110,7 +110,7 @@ namespace eosio { namespace cdt {
 
       void add_tuple(const clang::QualType& type) {
          auto pt = llvm::dyn_cast<clang::ElaboratedType>(type.getTypePtr());
-         auto tst = llvm::dyn_cast<clang::TemplateSpecializationType>(pt->desugar().getTypePtr());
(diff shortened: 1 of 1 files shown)


jeffreyssmith2nd:

PR #711, opened 2019-10-17
CI/CD support for Catalina

NorseGaud:
## Change Description

Added new Catalina 10.15.0 to the Buildkite CI pipelines.
Added SKIP_MACOS_10_15/14 to match what we use for EOSIO.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -67,7 +67,7 @@ steps:
           wait-network: true
     agents:
       - "queue=mac-anka-large-node-fleet"
-    skip: $SKIP_MOJAVE
+    skip: $SKIP_MACOS_10_14
 
   - label: ":darwin: macOS 10.15 - Build"
     command:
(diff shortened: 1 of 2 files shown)


.cicd/pipeline.yml
```diff
@@ -69,6 +69,30 @@ steps:
...
+          vm-name: 10.15.0_6C_14G_40G
+          vm-registry-tag: "clean::cicd::git-ssh::nas::brew::buildkite-agent"
+          modify-cpu: 12
+          modify-ram: 24
+          always-pull: true
+          debug: true
+          wait-network: true
+    agents:
+      - "queue=mac-anka-large-node-fleet"
+    skip: $SKIP_MOJAVE
```
jeffreyssmith2nd: Should these be `SKIP_CATALINA`? Or maybe `SKIP_MAC` if the intention is actually to skip all mac builds.

NorseGaud: Good catch. In fact, we need to change it completely to match what EOSIO uses. I've made the updates.

PR #735, opened 2019-11-07
Forward arguments to subprocesses correctly.  Fixes #531.

### LINKED ISSUE #531: v1.6.1 -D passing quotes properly?
I have (for testing purposes) the following code:
```c++
[[eosio::action]]
void t() {   
    print("\n[", __TEST__, "]");
}
```
if I compile the code like this:
```bash
$ eosio-cpp t.cpp -D__TEST__=worked
```
it will fail because it generates a line like this, which obviously needs quotes:
```
#define __TEST__ worked
```
so at first I tried:
```bash
$ eosio-cpp t.cpp -D__TEST__="worked"
```
but that didn't work:
```
#define __TEST__ worked
```
so I tried all these variations:
```bash
$ eosio-cpp t.cpp -D__TEST__=\"worked\"
#define __TEST__ worked
$ eosio-cpp t.cpp -D__TEST__="\""worked"\""
#define __TEST__ worked
$ eosio-cpp t.cpp -D__TEST__=\"\"worked\"\"
#define __TEST__ ""worked""
$ eosio-cpp t.cpp -D'__TEST__="worked"'
#define __TEST__ worked
$ eosio-cpp t.cpp -D'__TEST__=\"worked\"'
#define __TEST__ \"worked\"
$ eosio-cpp t.cpp -D"__TEST__=\"worked\""
#define __TEST__ worked
$ eosio-cpp t.cpp -D__TEST__=""worked""
#define __TEST__ worked
$ eosio-cpp t.cpp -D'__TEST__=""worked""'
#define __TEST__ ""worked""
```
the last of which puzzled me.  so 2 quotes inside of ticks come across but a single quote inside of ticks doesn't??

I'm reporting this as a bug (instead of asking on StackExchange) because I believe the preprocessor is doing something funky with quotes.  this is not a shell issue.

what I need is:
```
#define __TEST__ "worked"
```
can anyone figure out how to make that happen?

p.s. of course, I can't just write:
```c++
[[eosio::action]]
void t() {   
    print("\n[__TEST__]");
}
```
since the preprocessor will not replace values within string literals, so I need the define generated to quote the value

swatanabe-b1:
## Change Description

Concatenating the arguments and passing them to std::system doesn't handle quoted arguments correctly.  llvm already has a utility to handle this.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/cc/eosio-cpp.cpp.in
+++ b/tools/cc/eosio-cpp.cpp.in
@@ -201,7 +201,7 @@ int main(int argc, const char **argv) {
             output = opts.output_fn.empty() ? "a.out" : opts.output_fn;
          }
 
-         new_opts.insert(new_opts.begin(), "-o "+output);
+         new_opts.insert(new_opts.begin(), {"-o", output});
          outputs.push_back(output);
 
          if (llvm::sys::path::extension(input).equals(".c"))
@@ -221,7 +221,7 @@ int main(int argc, const char **argv) {
    if (opts.link) {
       std::vector<std::string> new_opts = opts.ld_options;
       for (auto input : outputs) {
-         new_opts.insert(new_opts.begin(), std::string(" ")+input+" ");
+         new_opts.insert(new_opts.begin(), input);
       }
    
       if (!eosio::cdt::environment::exec_subprogram("eosio-ld", new_opts)) {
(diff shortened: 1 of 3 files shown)


swatanabe-b1:
This is included with fixes in #736.

PR #748, opened 2019-11-21
Track ricardian file dependencies. 

### LINKED ISSUE #309: CMake ricardian clauses not re-generated on resource files changes
### Versions:
 * eosio.cdt 1.4.1

### Description
When using `CMake`, if the actual resource file containing the ricardian clauses (either the contracts or clauses file) is the only changes, the ABI file is not re-generated correctly when doing a simple `make`.

Repro:

```
# Assuming at root of this repository
cd `examples/hello`
mkdir build
cd build && cmake .. && make && cd ..
echo "Adding stuff at the end of clauses" >> hello.clauses.md
cd build && make cd ..
cat build/hello.abi | grep "Adding stuff at the end of clauses"
echo $?
# Prints 1, should be 0
```

Same thing if `echo "Adding stuff at the end of clauses" >> hello.contracts.md`. The `grep` exit code should be 0.

Sorry if this is already tracked, did a search but did not find anything related.

### LINKED ISSUE #455: `target_ricardian_directory` macro fails to find target
``` cmake
   macro (target_ricardian_directory TARGET DIR)
      target_compile_options( ${TARGET} PUBLIC -R${DIR} )
      file(GLOB contracts ${DIR}/*.contracts.md ${DIR}/*.clauses.md)
      add_dependencies( ${TARGET} ${contracts} )
   endmacro()
```

This macro calls `add_dependencies` to add ricardian files to the contract, but cmake complains non-existence of target. `add_dependencies` creates a relationship between top-level targets, but contract is an external project, so it seems that contract is not visible from top-level execution. Without cmake version, contract is built regardless of this error, but when `cmake_minimum_required` is added with the latest version, build process is aborted.

swatanabe-b1:
## Change Description

add_dependencies does not for for ordinary files.
Use OBJECT_DEPENDS instead and also use the current source directory by default.
If target_ricardian_directory is called, it will override this default.
Fixes #455.  Fixes #309.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


jeffreyssmith2nd:
This is causing CI to fail with 

```

CMake Error at /Users/anka/eosio.cdt/build/lib/cmake/eosio.cdt/EosioCDTMacros.cmake:9 (set_source_files_properties):
--
  | set_source_files_properties called with incorrect number of arguments.

```

PR #768, opened 2019-12-19
Docs as Code for source installation + Pipeline generation

NorseGaud:
## Change Description

**Problem:** Source install documentation is typically updated last, after changes to CI and EOSIO have happened. Multiple teams are involved too, and communication can fail or get unclear.

**Solution:** Refactor CI scripts and pipelines to pull commands from documentation so we ensure updates are made to the docs before CI can pass. This causes all teams involved to update the docs, instead of only Developer Relations.

What these changes do:

1. Generation of tags (docker & anka) are done using every code block EXCEPT for the TEST block.
    - Why? We use CTEST and not `make test` in our scripts so we can split up the tests into individual jobs/steps.
    - The tag will ONLY push to registries if commands in the entire document function without generating a failure. This is an improvement over how we currently generate tags since we only run the dependency install and don't test if they actually work or not. We can avoid filling up the registries with bad images now.
2. Execution of build, install, and uninstall commands using commands in md files.
3. $HOME/eosio is the new root. The commands will clone into $HOME/eosio/eos. All dependencies and eos binaries will end up in $HOME/eosio/install/(bin|lib|etc). 
    - If users want to have more than one eosio installation on their system (common for BPs), they can modify the EOSIO_LOCATION env.
    - CDT and Contracts will also use $HOME/eosio/cdt & $HOME/eosio/install, etc. This allows the creation of a new build/install script that pulls the commands from documentation for eos, cdt, contracts, and executes them in order for the user. The script will only need to be updated with compatible version combinations so it can find the proper documentation and commands.

## Code block rules/conversions:

- The following are removed from all code blocks: `grep -v -e "$PATTERN" -e '<!--' -e '-->' -e '```' -e '\#.*' -e '^$'`
- All code blocks in the MD files are wrapped with a comment like `<!-- DAC CLONE -->` or `<!-- DAC BUILD -->`, allowing our scripts to pull specific sections.
- Anything inside of `<!-- DAC IGNORE` comments are ignored (like the test block)
- Indented lines are executed with the last non-indented line. This allows you to execute multiple commands in the same docker layer.

#### Dockerfile
1. Lines matching `^[ ].*` are printed without any modification.
2. Lines starting with `^export.*=([^\s]+)$` are converted to have ENV before them.

#### Mac SH script
1. Nothing :)
---

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


jeffreyssmith2nd:
Decided to go a different direction with this

PR #771, opened 2020-01-02
Separate write_decimal from eosiolib

### LINKED ISSUE #770: v1.6.3 breaks leagacy-server.wasm in history-tools
`write_decimal` was introduced in v1.6.3 and a linkage to `eosiolib.cpp.obj` is enforced when using `eosio::asset`. This adds unresolvable imports like (current_time, set_blockchain_parameters_packed, get_blockchain_parameters_packed, set_proposed_producers, get_active_producers) to legacy-server.wasm of history-tools, so wasm-ql engine fails to load it.

conr2d:
## Change Description
Fixes #770. 

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/core/eosio/asset.hpp
+++ b/libraries/eosiolib/core/eosio/asset.hpp
@@ -4,13 +4,12 @@
 #include "print.hpp"
 #include "check.hpp"
 #include "symbol.hpp"
+#include "stringify.hpp"
 
 #include <tuple>
 #include <limits>
 
 namespace eosio {
-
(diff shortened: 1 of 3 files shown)


swatanabe-b1:
This will cause a link error if stringify.hpp is included in more than one translation unit.

conr2d:
@larryk85 
Can you refer to #770? Using `eosio::asset` enforces linking to eosiolib.cpp.obj (libraries/eosiolib/eosiolib.cpp) due to `write_decimal` and imports like (set_blockchain_parameters_packed and so on) are added to binary. However, binaries for wasmql in history-tools should not have them.

conr2d:
Replaced by #775.

PR #797, opened 2020-01-24
Add Key Value api for the new key value intrinsics.

jeffreyssmith2nd:
## Change Description

Introduces a new api for interacting with the key value intrinsics.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


Doxyfile
```diff
@@ -1980,7 +1980,7 @@ INCLUDE_FILE_PATTERNS  =
 # recursively expanded use the := operator instead of the = operator.
 # This tag requires that the tag ENABLE_PREPROCESSING is set to YES.
 
-PREDEFINED             =
+PREDEFINED             = _DOXYGEN_
```
larryk85: C++ spec 17.4.3.1.2. These types of names are reserved for the compiler.  I would simply remove the underscores and make the defined name a bit more specific so it doesn't clash.  So something like `EOSIO_CDT_DOXYGEN`.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1055 @@
...
+
+         __attribute__((eosio_wasm_import))
+         int32_t kv_it_value(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
+      }
+   }
+
+/**
+ * The key_type struct is used to store the binary representation of a key.
+ */
+struct key_type {
```
larryk85: This type can simply inherit from string and retain type safety so conversion from a string can't be made implicitly.  We can also inherit from eosio::string and not std::string to save on WASM bloat generated from std::string.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1055 @@
...
+   bool operator==(const key_type& k) const {
+      return std::tie(size, buffer) == std::tie(k.size, k.buffer);
+   }
+   bool operator!=(const key_type& k) const {
+      return !(*this == k);
+   }
+};
+
+namespace detail {
+   constexpr static size_t max_stack_buffer_size = 512;
```
larryk85: instead of static, I would use `inline` for these types of global variables.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: After the `Release` fix, I encountered a max call depth error. Switching to std::string leads to no errors, as well as no noticeable binary bloat, so I went with that.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+
+            if (idx->name == idx->tbl->primary_index->name) {
+               datastream<const char*> ds((char*)buffer, actual_value_size);
+
+               T val;
+               ds >> val;
+               if (data_size > detail::max_stack_buffer_size) {
+                  free(buffer);
+               }
+               return val;
```
swatanabe-b1: This is duplicated below.  You can reduce the duplication by pulling the extra lookup above this and then making the deserialization unconditional.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+            eosio::check(itr_stat != kv_it_stat::iterator_end, "cannot increment end iterator");
+            itr_stat = static_cast<kv_it_stat>(internal_use_do_not_use::kv_it_next(itr));
+            return *this;
+         }
+
+         iterator& operator--() {
+            itr_stat = static_cast<kv_it_stat>(internal_use_do_not_use::kv_it_prev(itr));
+            eosio::check(itr_stat != kv_it_stat::iterator_end, "incremented past the beginning");
+            return *this;
+         }
```
swatanabe-b1: You should also provide `operator++(int)` and `operator--(int)`

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+            uint32_t actual_value_size;
+            uint32_t offset = 0;
+
+            void* buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
+            auto stat = internal_use_do_not_use::kv_it_value(itr, offset, (char*)buffer, data_size, actual_value_size);
+
+            eosio::check(static_cast<kv_it_stat>(stat) != kv_it_stat::iterator_end, "Error reading value");
+
+            if (idx->name == idx->tbl->primary_index->name) {
+               datastream<const char*> ds((char*)buffer, actual_value_size);
```
swatanabe-b1: If actual_value_size > data_size (which can happen if the value was modified after data_size was last updated), this will access garbage memory.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+
+         iterator& operator++() {
+            eosio::check(itr_stat != kv_it_stat::iterator_end, "cannot increment end iterator");
+            itr_stat = static_cast<kv_it_stat>(internal_use_do_not_use::kv_it_next(itr));
+            return *this;
+         }
+
+         iterator& operator--() {
+            itr_stat = static_cast<kv_it_stat>(internal_use_do_not_use::kv_it_prev(itr));
+            eosio::check(itr_stat != kv_it_stat::iterator_end, "incremented past the beginning");
```
swatanabe-b1: Did you mean decremented?

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+       *       auto itr = addresses.index.account_name.find("dan"_n);
+       *       eosio::check(itr != addresses.index.account_name.end(), "Couldn't get him.");
+       *
+       *       auto itr = addresses.index.full_name.find("Dan Larimer");
+       *       eosio::check(itr != addresses.index.full_name.end(), "Couldn't get him.");
+       *     }
+       * @endcode
+       */
+      template <typename K>
+      iterator find(K key) {
```
swatanabe-b1: Pass by reference.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+         if (!success) {
+            return end();
+         }
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(db, contract_name.value, prefix.data(), prefix.size());
+
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+         auto cmp = internal_use_do_not_use::kv_it_key_compare(itr, t_key.data(), t_key.size());
+
+         eosio::check(cmp == 0, "This key does not exist in this iterator");
```
swatanabe-b1: What purpose does this comparison serve?  You already checked that the key exists using get and the table could not have changed since then.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+       */
+      iterator begin() {
+         auto prefix = make_prefix(table_name, name);
+         uint32_t itr = internal_use_do_not_use::kv_it_create(db, contract_name.value, prefix.data(), prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, "", 0);
+
+         uint32_t value_size;
+
+         // kv_it_value is just used to get the value_size
+         void* buffer = alloca(1);
```
swatanabe-b1: `char buffer[1]`
Actually, nullptr works just fine.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+    *       void myaction() {
+    *          auto t = mytable::open("mycontract"_n);
+    *       }
+    * }
+    *
+    * @endcode
+    */
+   static Class open(eosio::name contract_name) {
+      Class c = Class(contract_name);
+      return c;
```
swatanabe-b1: `return Class(contract_name)`

libraries/eosiolib/core/eosio/utility.hpp
```diff
@@ -0,0 +1,40 @@
...
+         union i64_swap {
+            T value;
+            uint64_t ui;
+            i64_swap(T v) : value(v) { ui = __builtin_bswap64(ui); }
+         };
+         return i64_swap(val).value;
+      } else {
+         union i128_swap {
+            uint64_t high;
+            uint64_t low;
```
swatanabe-b1: high and low are the same, because they're members of the same union.

libraries/eosiolib/core/eosio/utility.hpp
```diff
@@ -0,0 +1,40 @@
...
+   /* utility function to swap the endianness of a variable */
+   template <typename T>
+   inline T swap_endian(T val) {
+      static_assert(sizeof(T) <= 16);
+      if constexpr (sizeof(T) == 1)
+         return val;
+      else if constexpr (sizeof(T) == 2)
+         return (val << 8) |(val >> 8);
+      else if constexpr (sizeof(T) == 4) {
+         union i32_swap {
```
swatanabe-b1: Using a union like this may have undefined behavior depending on T.  signed int may alias unsigned int, but not long even if int and long have the same size. Use memcpy instead.

tests/unit/test_contracts/kv_make_key_tests.cpp
```diff
@@ -0,0 +1,296 @@
...
+      ++itr;
+      eosio::check(itr.value().tname == s5.tname, "Got the wrong value");
+      ++itr;
+      eosio::check(itr.value().tname == s1.tname, "Got the wrong value");
+      ++itr;
+      eosio::check(itr.value().tname == s4.tname, "Got the wrong value");
+      ++itr;
+      eosio::check(itr.value().tname == s3.tname, "Got the wrong value");
+      ++itr;
+      eosio::check(itr == end_itr, "Should be the end");
```
swatanabe-b1: This is very repetitive.  If you compare the entire struct instead of just the specific key, you can define a helper function that can be called like:
```
check_index(t.index.tname, {s2, s5, s1, s4, s3});
```

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+
+   if (case_insensitive) {
+      std::transform(str, str + size, (char*)data_buffer, [](unsigned char c) -> unsigned char { return std::toupper(c); });
+   } else {
+      memcpy(data_buffer, str, size);
+   }
+
+   ((char*)data_buffer)[data_size - 3] = 0x01;
+   ((char*)data_buffer)[data_size - 2] = 0x00;
+   ((char*)data_buffer)[data_size - 1] = 0x00;
```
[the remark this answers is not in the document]
jeffreyssmith2nd: This was a misunderstanding of the encoding needed for strings to be properly comparable. I'm currently working on the correct version based on feedback from Todd.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: I have commit coming that removes data_size completely and gets the size right before these operations. I think that will address this concern.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+   key_type s((const char*)buffer, buffer_size);
+
+   if (buffer_size > detail::max_stack_buffer_size) {
+      free(buffer);
+   }
+
+   return s;
+}
+
+inline key_type table_key(const key_type& prefix, const key_type& key) {
```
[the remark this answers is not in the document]
jeffreyssmith2nd: Good catch, thanks.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: I restructured this code to remove the get check and only use kv_it_key_compare instead since my understanding of the code is that will be more performant.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,874 @@
...
+         __attribute__((eosio_wasm_import))
+         int32_t kv_it_value(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
+      }
+   }
+
+/**
+ * The key_type struct is used to store the binary representation of a key.
+ */
+struct key_type : std::string {
+   using std::string::string;
```
tbfleming: I believe this enables unwanted implicit conversions.

tests/unit/test_contracts/kv_make_key_tests.cpp
jeffreyssmith2nd: This cleaned it up a lot, thanks.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+
+#include <algorithm>
+#include <cctype>
+#include <functional>
+
+#include <boost/preprocessor/variadic/to_seq.hpp>
+#include <boost/preprocessor/seq/for_each_i.hpp>
+#include <boost/pfr.hpp>
+
+#define EOSIO_CDT_KV_INDEX_nullptr
```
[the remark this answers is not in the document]
jeffreyssmith2nd: TEST_1 causes a compilation error, though I'm not sure how to resolve that. TEST works fine. I'm not sure exactly which `_` you're referring to, nothing is being prefixed with a `_` so we shouldn't run into an `__` names.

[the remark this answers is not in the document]
swatanabe-b1: Below you have `EOSIO_CDT_KV_INDEX_ ## index_name`  This must not expand to either:
a) a reserved identifier  (any name containing a double underscore is reserved to the implementation for any use)
b) a macro that you defined that it isn't intended to match.
Since `index_name` is user-defined, the only constraint on it is that it must not be a reserved name.  The fix for both is to change the prefix `EOSIO_CDT_KV_INDEX_`

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,870 @@
...
+}
+
+template <typename S, typename std::enable_if_t<std::is_class<S>::value, int> = 0>
+inline key_type make_key(S val) {
+   size_t data_size = 0;
+   size_t pos = 0;
+   void* data_buffer;
+
+   boost::pfr::for_each_field(val, [&](auto& field) {
+      data_size += sizeof(field);
```
tbfleming: Doesn't work for most member types: string, vector, ...

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+   }
+   return s;
+}
+
+inline key_type make_key(const std::string& val, bool case_insensitive=false) {
+   return make_key(val.data(), val.size(), case_insensitive);
+}
+
+inline key_type make_key(const char* str, bool case_insensitive=false) {
+   return make_key(std::string{str}, case_insensitive);
```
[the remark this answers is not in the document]
jeffreyssmith2nd: I was using std::string here to find the length for us, are you recommending its worth avoiding the copy and having to implement the length finding ourselves? Or is there a better way to find the length?

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+       * Returns an iterator referring to the `past-the-end` element. It does not point to any element, therefore `value` should not be called on it.
+       * @ingroup keyvalue
+       *
+       * @return An iterator referring to the `past-the-end` element.
+       */
+      iterator end() {
+         auto prefix = make_prefix(table_name, name);
+         uint32_t itr = internal_use_do_not_use::kv_it_create(db, contract_name.value, prefix.data(), prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_move_to_end(itr);
+
```
tbfleming: Perfomance issues:
* `kv_it_create` creates an iterator at the end. `kv_it_move_to_end` is redundant.
* The following pattern creates and destroys intrinsic iterators each iteration through the loop:

```
for(auto it = x.begin(); it != x.end(); ++it)
```

A potential fix: `end()` could return a `kv_index::iterator` with a null intrinsic iterator.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+    * @ingroup keyvalue
+    *
+    * @brief Defines an index on an EOSIO Key Value Table
+    * @details A Key Value Index allows a user of the table to search based on a given field.
+    * The only restrictions on that field are that it is serializable to a binary representation.
+    * Convenience functions exist to handle most of the primitive types as well as some more complex types, and are
+    * used automatically where possible.
+    */
+   class kv_index {
+      class iterator {
```
tbfleming: Need move-construct and move-assignment

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
swatanabe-b1: That's why I suggested string_view, which finds the length without making a copy.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+         auto prefix = make_prefix(table_name, name);
+         auto t_key = table_key(prefix, make_key(key));
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(db, contract_name.value, prefix.data(), prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+
+         auto cmp = internal_use_do_not_use::kv_it_key_compare(itr, t_key.data(), t_key.size());
+
+         if (cmp != 0) {
+            return end();
```
tbfleming: This leaks the intrinsic iterator

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+      template <typename K>
+      iterator lower_bound(K key) {
+         auto prefix = make_prefix(table_name, name);
+         auto t_key = table_key(prefix, make_key(key));
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(db, contract_name.value, prefix.data(), prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+
+         if (static_cast<kv_it_stat>(itr_stat) == kv_it_stat::iterator_end) {
+            return end();
```
tbfleming: leak

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+      }
+
+      /**
+       * Returns an iterator referring to the `past-the-end` element. It does not point to any element, therefore `value` should not be called on it.
+       * @ingroup keyvalue
+       *
+       * @return An iterator referring to the `past-the-end` element.
+       */
+      iterator end() {
+         auto prefix = make_prefix(table_name, name);
```
tbfleming: Performance: make prefix a kv_index member

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+            return std::nullopt;
+         }
+
+         void* buffer = value_size > detail::max_stack_buffer_size ? malloc(value_size) : alloca(value_size);
+         auto copy_size = internal_use_do_not_use::kv_get_data(db, 0, (char*)buffer, value_size);
+
+         datastream<const char*> ds((char*)buffer, value_size);
+
+         T val;
+         ds >> val;
```
tbfleming: Will malfunction on secondary index

tbfleming: performance:
* have a local `std::optional<T>` variable
* always return this variable for success or failure
* emplace then deserialize it

tbfleming: That will enable return-value optimization

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+
+         if (static_cast<kv_it_stat>(itr_stat) == kv_it_stat::iterator_end) {
+            return end();
+         }
+
+         return {contract_name, itr, static_cast<kv_it_stat>(itr_stat), this};
+      }
+
+      /**
+       * Returns an iterator pointing to the element with the highest key less than or equal to the given key.
```
tbfleming: This doesn't match standard behavior. See https://en.cppreference.com/w/cpp/algorithm/upper_bound

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+       * @ingroup keyvalue
+       *
+       * @tparam K - The type of the key. This will be auto-deduced by the key param.
+       *
+       * @param begin - The beginning of the range.
+       * @param end - The end of the range.
+       * @return A vector containing all the objects that fall between the range.
+       */
+      template <typename K>
+      std::vector<T> range(K begin, K end) {
```
tbfleming: This logic can be much simpler once upper_bound changes to match std::upper_bound's definition

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+   class null_kv_index : public kv_index {};
+
+   /**
+    * Puts a value into the table. If the value already exists, it updates the existing entry.
+    * The key is determined from the defined primary index.
+    * @ingroup keyvalue
+    *
+    * @param value - The entry to be stored in the table.
+    */
+   void put(const T& value) {
```
tbfleming: This doesn't clean up secondary indexes if the value already exists and secondary keys changed

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: The way I understand that is to have (abbreviated):
```
std::optional<T> ret_val;
...
ret_val.emplace();
ds >> ret_val;
...
return ret_val;
```

The problem I'm running into with that is datastream assumes an optional was serialized into it, however when we store a value its not an optional. So deserialization fails.

swatanabe-b1: `ds >> *ret_val`

jeffreyssmith2nd: Thanks 👍

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+            eosio::check(itr_stat != kv_it_stat::iterator_end, "Cannot read end iterator");
+
+            uint32_t value_size;
+            uint32_t actual_value_size;
+            uint32_t offset = 0;
+
+            // call once to get the value_size
+            internal_use_do_not_use::kv_it_value(itr, 0, (char*)nullptr, 0, value_size);
+
+            eosio::check(value_size > 0, "Cannot read a value of size 0");
```
tbfleming: Why not?

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,863 @@
...
+
+            if (idx->name != idx->tbl->primary_index->name) {
+               uint32_t actual_data_size;
+               auto success = internal_use_do_not_use::kv_get(db, contract_name.value, (char*)buffer, actual_value_size, actual_data_size);
+               eosio::check(success, "failure getting primary key");
+
+               void* pk_buffer = actual_data_size > detail::max_stack_buffer_size ? malloc(actual_data_size) : alloca(actual_data_size);
+               auto copy_size = internal_use_do_not_use::kv_get_data(db, 0, (char*)pk_buffer, actual_data_size);
+
+               eosio::check(copy_size > 0, "failure getting primary index data");
```
tbfleming: Why disallow empty?

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: Bucky and I decided intentionally not to, so that people don't inadvertently use them and suffer the copy penalty.

tbfleming: Plus the intrinsics don't support copying iterators

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: Is decrementing an end iterator considered valid? Or is an end iterator only ever allowed to point off the end?

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: For the purposes of serialization/deserialization, what would a size 0 value actually mean? Don't all types have at least size 1?

libraries/eosiolib/contracts/eosio/key_value.hpp
tbfleming: C++: if begin != end, then decrementing end is valid

tbfleming: kv intrinsics: incrementing and decrementing end is always valid. If nothing matches the prefix, then the result is end (circles back on itself).

jeffreyssmith2nd: Okay so if we had end store a null iterator, we would need to create an iterator inside the `--` operator if one didn't already exist.

libraries/eosiolib/contracts/eosio/key_value.hpp
tbfleming: An empty struct serializes to size 0. This is not the sizeof operator.

libraries/eosiolib/contracts/eosio/key_value.hpp
tbfleming: yes

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: The best solution I could come up with is just deleting all secondary indexes before `set`ting them because I couldn't think of a way to derive the old value of the index (if it has changed) with the context we have here.

I'm open to a better solution if there is one.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: This will be addressed once the abieos refactor is complete.

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
jeffreyssmith2nd: This will be addressed once the abieos refactor is complete.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+   constexpr size_t size = sizeof(big_endian);
+   key_type s(bytes, size);
+   return s;
+}
+
+template <typename I, typename F>
+inline key_type make_floating_key(F val) {
+   auto* ival = reinterpret_cast<I*>(&val);
+   I bit_val;
+   auto msb = get_msb(*ival);
```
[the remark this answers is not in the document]
jeffreyssmith2nd: This will be addressed once the abieos refactor is complete.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: This PR was getting large, so I added the changes for this in a new PR https://github.com/EOSIO/eosio.cdt/pull/807 that goes into this branch.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,880 @@
...
+
+         ~iterator() {
+            if (itr) {
+               internal_use_do_not_use::kv_it_destroy(itr);
+            }
+         }
+
+         iterator& operator=(iterator&& other) {
+            contract_name = std::move(other.contract_name);
+            itr = std::exchange(other.itr, 0);
```
swatanabe-b1: Leaks the original value of the iterator.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,880 @@
...
+
+            if (idx->name != idx->tbl->primary_index->name) {
+               uint32_t actual_data_size;
+               auto success = internal_use_do_not_use::kv_get(idx->tbl->db_name, contract_name.value, (char*)buffer, actual_value_size, actual_data_size);
+               eosio::check(success, "failure getting primary key");
+
+               void* pk_buffer = actual_data_size > detail::max_stack_buffer_size ? malloc(actual_data_size) : alloca(actual_data_size);
+               auto copy_size = internal_use_do_not_use::kv_get_data(idx->tbl->db_name, 0, (char*)pk_buffer, actual_data_size);
+
+               eosio::check(copy_size != actual_value_size, "failure getting primary index data");
```
swatanabe-b1: I'm not sure what you're trying to test here.  Won't this fail when the primary key has the same size as its value?

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,880 @@
...
+       * Get the value for an existing object in a table by the index, using the given key.
+       * @ingroup keyvalue
+       *
+       * @tparam K - The type of the key. This will be auto-deduced by the key param.
+       *
+       * @param key - The key to search for.
+       * @return A std::optional of the value corresponding to the key.
+       */
+      template <typename K>
+      std::optional<T> get(K&& key) {
```
swatanabe-b1: This seems somewhat error-prone, as there is nothing to enforce that Key has the correct type.  e.g. passing a literal 0 for a uint64_t index.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: Yeah I think this is an artifact of old code (that looks like it might have been wrong originally anyways). I don't think the intent of this test is really needed.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: I definitely agree. I haven't been able to come up with a solution yet though that allows us to template the kv_index on the key type in a way that can be deduced so that we don't have to specify it.

swatanabe-b1: Have you tried CTAD

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1051 @@
...
+      return make_floating_key<uint64_t>(val);
+   }
+}
+
+inline key_type make_key(const char* str, size_t size, bool case_insensitive=false) {
+   size_t data_size = size + 3;
+   void* data_buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
+
+   if (case_insensitive) {
+      std::transform(str, str + size, (char*)data_buffer, [](unsigned char c) -> unsigned char { return std::toupper(c); });
```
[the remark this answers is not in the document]
jeffreyssmith2nd: As discussed previously in person, we're not going to attempt unicode support for toupper.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1075 @@
...
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(tbl->db_name, contract_name.value, prefix.data(), prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+
+         iterator it{contract_name, itr, static_cast<kv_it_stat>(itr_stat), this};
+
+         auto cmp = internal_use_do_not_use::kv_it_key_compare(it.itr, t_key.data(), t_key.size());
+         if (cmp == 0) {
+            ++it;
+         }
```
swatanabe-b1: This doesn't handle non-unique keys correctly.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1075 @@
...
+
+            void* buffer = value_size > detail::max_stack_buffer_size ? malloc(value_size) : alloca(value_size);
+            auto stat = internal_use_do_not_use::kv_it_value(itr, offset, (char*)buffer, value_size, actual_value_size);
+
+            eosio::check(static_cast<kv_it_stat>(stat) != kv_it_stat::iterator_end, "Error reading value");
+
+            void* deserialize_buffer = buffer;
+            size_t deserialize_size = actual_value_size;
+
+            if (idx->name != idx->tbl->primary_index->name) {
```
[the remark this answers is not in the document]
jeffreyssmith2nd: There currently isn't any equality operator defined for `kv_index`, so I don't think that would work.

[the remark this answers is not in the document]
swatanabe-b1: idx is a pointer.  equality is defined for pointers.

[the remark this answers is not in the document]
jeffreyssmith2nd: Ah right, thanks.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+         if (!itr) {
+            itr = internal_use_do_not_use::kv_it_create(config.db_name, config.contract_name.value, config.prefix.data(), config.prefix.size());
+         }
+         itr_stat = static_cast<kv_it_stat>(internal_use_do_not_use::kv_it_prev(itr));
+         eosio::check(itr_stat != kv_it_stat::iterator_end, "decremented past the beginning");
+         return *this;
+      }
+
+      uint32_t key_compare(key_type kt) const {
+         return internal_use_do_not_use::kv_it_key_compare(itr, kt.data(), kt.size());
```
swatanabe-b1: `kv_it_compare` returns int32_t.

swatanabe-b1: Also, this behaves differently when ``itr == 0`` from when ``itr`` is an end iterator.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+      using kv_table<T>::kv_index::tbl;
+      using kv_table<T>::kv_index::table_name;
+      using kv_table<T>::kv_index::contract_name;
+      using kv_table<T>::kv_index::index_name;
+      using kv_table<T>::kv_index::prefix;
+
+      index() = default;
+
+      template <typename KF>
+      index(KF&& kf) : kv_index{kf} {
```
swatanabe-b1: ``explicit``

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+         }
+
+         return {itr, static_cast<kv_it_stat>(itr_stat), config};
+      }
+
+      /**
+       * Returns an iterator pointing to the first element greater than the given key.
+       * @ingroup keyvalue
+       *
+       * @return An iterator pointing to the element with the highest key less than or equal to the given key.
```
swatanabe-b1: Nope.  The first description is correct.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+       *
+       * @return An iterator pointing to the element with the lowest key greater than or equal to the given key.
+       */
+      iterator lower_bound(const K& key) {
+         auto t_key = table_key(config.prefix, make_key(key));
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(config.db_name, config.contract_name.value, config.prefix.data(), config.prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+
+         if (static_cast<kv_it_stat>(itr_stat) == kv_it_stat::iterator_end) {
```
swatanabe-b1: Is this test useful?  The iterator is already at end.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+       *
+       * @return An iterator pointing to the element with the highest key less than or equal to the given key.
+       */
+      iterator upper_bound(const K& key) {
+         auto t_key = table_key(config.prefix, make_key(key));
+
+         uint32_t itr = internal_use_do_not_use::kv_it_create(config.db_name, config.contract_name.value, config.prefix.data(), config.prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, t_key.data(), t_key.size());
+
+         iterator it{itr, static_cast<kv_it_stat>(itr_stat), config};
```
swatanabe-b1: This is (almost) the same as lower_bound.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+            return {};
+         }
+
+         std::vector<T> return_values;
+
+         iterator itr = std::move(begin_itr);
+         while(itr < end_itr) {
+            return_values.push_back(itr.value());
+            ++itr;
+         }
```
swatanabe-b1: This can be reduced to an old-style for loop:
```c++
for(auto itr = lower_bound(b), end_iter = lower_bound(e); itr < end_iter; ++iter)
```

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+               free(buffer);
+            }
+         }
+
+         if (primary_key_found) {
+            auto old_sec_key = table_key(make_prefix(table_name, idx->index_name), idx->get_key(old_value));
+            internal_use_do_not_use::kv_erase(db_name, contract_name.value, old_sec_key.data(), old_sec_key.size());
+         }
+
+         internal_use_do_not_use::kv_set(db_name, contract_name.value, sec_tbl_key.data(), sec_tbl_key.size(), tbl_key.data(), tbl_key.size());
```
swatanabe-b1: If the secondary key is not modified (probably the common case), this causes an unnecessary erase+set.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: In the case where `itr==0` and it's not an end iterator, it should be fine just to return some negative number to signal `less than` right?

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: Yeah, looking at it now I don't think it actually adds any value.

libraries/eosiolib/contracts/eosio/key_value.hpp
swatanabe-b1: > itr==0 and it's not an end iterator

Impossible.  ``itr == 0`` is an end iterator, from the wrapper's perspective.

jeffreyssmith2nd: Then I'm not sure I understand what you mean by
> Also, this behaves differently when itr == 0 from when itr is an end iterator.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+       */
+      iterator begin() {
+         uint32_t itr = internal_use_do_not_use::kv_it_create(config.db_name, config.contract_name.value, config.prefix.data(), config.prefix.size());
+         int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, "", 0);
+
+         return {itr, static_cast<kv_it_stat>(itr_stat), config};
+      }
+
+      /**
+       * Returns an iterator referring to the `past-the-end` element. It does not point to any element, therefore `value` should not be called on it.
```
tbfleming: There is no `past-the-end` element. Instead the iterator is past the end.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+inline key_type make_key(const char* str, bool case_insensitive=false) {
+   return make_key(std::string_view{str}, case_insensitive);
+}
+
+inline key_type make_key(eosio::name n) {
+   return make_key(n.value);
+}
+
+inline key_type make_key(key_type&& val) {
+   return val;
```
swatanabe-b1: std::move

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+      index_config config;
+
+   public:
+      using kv_table<T>::kv_index::tbl;
+      using kv_table<T>::kv_index::table_name;
+      using kv_table<T>::kv_index::contract_name;
+      using kv_table<T>::kv_index::index_name;
+      using kv_table<T>::kv_index::prefix;
+
+      index() = default;
```
swatanabe-b1: When would a default constructed index make sense?

libraries/eosiolib/contracts/eosio/key_value.hpp
swatanabe-b1: And now it's identical to lower_bound.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+            eosio::check(!sec_found, "Attempted to store an existing secondary index.");
+            internal_use_do_not_use::kv_set(db_name, contract_name.value, sec_tbl_key.data(), sec_tbl_key.size(), tbl_key.data(), tbl_key.size());
+         } else {
+            if (sec_found) {
+               void* buffer = value_size > detail::max_stack_buffer_size ? malloc(value_size) : alloca(value_size);
+               auto copy_size = internal_use_do_not_use::kv_get_data(db_name, 0, (char*)buffer, value_size);
+
+               eosio::check(copy_size == tbl_key.size(), "Attempted to update an existing secondary index.");
+               auto res = memcmp(buffer, tbl_key.data(), copy_size);
+               eosio::check(res == 0, "Attempted to update an existing secondary index.");
```
swatanabe-b1: I think it would be better to combine the two checks.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+         index->index_name = eosio::name{index_name};
+      }
+
+      index->contract_name = contract_name;
+      index->table_name = table_name;
+      index->tbl = this;
+
+      index->setup();
+      secondary_indices.push_back(index);
+   }
```
swatanabe-b1: This duplicates a lot of code from the other overload.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+      if (primary_index->index_name.value > 0) {
+         is_named = true;
+      } else {
+         primary_index->index_name = eosio::name{index_name};
+         ++index_name;
+      }
+
+      primary_index->setup();
+
+      if constexpr (sizeof...(indices) > 0) {
```
swatanabe-b1: It would be cleaner to define an overload of setup_indices that takes no indices.  This would also simplify your recursion.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1005 @@
...
+         setup_indices(index_name, is_named, indices...);
+      }
+   }
+
+   template <typename I, typename... Indices>
+   void setup_indices(uint64_t index_name, bool is_named, I index, Indices... indices) {
+      if (is_named) {
+         eosio::check(index->index_name.value > 0, "All indices must be named if one is named.");
+      } else {
+         eosio::check(index->index_name.value <= 0, "All indices must be named if one is named.");
```
swatanabe-b1: This looks a bit odd, since name.value is unsigned and therefore cannot be less than 0.

tests/integration/kv_make_key_tests.cpp
```diff
@@ -0,0 +1,73 @@
...
+   t.set_abi(config::system_account_name, contracts::kv_bios_abi().data());
+
+   auto data = mvo()("k", 1024)("v", 1024*1024)("i", 256);
+   t.push_action(config::system_account_name, N(ramkvlimits), config::system_account_name, data);
+   t.push_action(N(kvtest), N(setup), N(kvtest), {});
+   t.push_action(N(kvtest), test_name, N(kvtest), {});
+}
+
+BOOST_AUTO_TEST_CASE(makekeyname) try {
+   make_key_test(N(makekeyname));
```
swatanabe-b1: All of these test cases could be a single BOOST_DATA_TEST_CASE.

tests/unit/test_contracts/kv_multiple_indices_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+   eosio::name primary_key;
+   std::string foo;
+   uint64_t bar;
+
+   std::string fullname;
+   uint32_t age;
+
+   std::tuple<std::string, uint32_t> non_unique_name() const { return {fullname, age}; }
+
+   bool operator==(const my_struct b) const {
```
swatanabe-b1: Should pass b by reference.

tests/unit/test_contracts/kv_multiple_indices_tests.cpp
```diff
@@ -0,0 +1,258 @@
...
+      auto bar_begin_itr = t.bar.begin();
+      auto bar_end_itr = t.bar.end();
+
+      auto foo_itr = t.foo.begin();
+      auto bar_itr = t.bar.begin();
+
+      eosio::check(foo_itr != foo_end_itr, "Should not be the end");
+      eosio::check(bar_itr != bar_end_itr, "Should not be the end");
+      eosio::check(foo_itr.value() == s2, "Got the wrong value");
+      eosio::check(bar_itr.value() == s5, "Got the wrong value");
```
swatanabe-b1: You have a lot of identical error messages, which will make it hard to find the source of a failure.  Consider reporting the line#.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -86,6 +86,20 @@
...
+ * @details In the case where the autogenerated index names created by DEFINE_TABLE are not enough, a user can instead
+ * manually define the table and indices. This macro allows users to conveniently define an index without having to specify
+ * the index template type, as those can be large/unwieldy to type out.
+ *
+ * @param variable_name - The variable name in code of the index.
+ * @param index_name    - The index name.
+ * @param value_class   - The class the table stores.
+ * @parma member_name   - The name of the member pointer used for the index.
+ */
+#define NAMED_INDEX(variable_name, index_name, value_class, member_name)                                               \
```
swatanabe-b1: You can avoid requiring value_class by creating a value_type typedef in kv_table or some such.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,993 @@
...
+ * The key_type struct is used to store the binary representation of a key.
+ */
+struct key_type : private std::string {
+   key_type() : std::string() {}
+   key_type(const char* c, size_t s) : std::string(c, s) {}
+
+
+   key_type operator+(const key_type& b) const {
+      size_t buffer_size = size() + b.size();
+      void* buffer = buffer_size > detail::max_stack_buffer_size ? malloc(buffer_size) : alloca(buffer_size);
```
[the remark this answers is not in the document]
swatanabe-b1: Undefined behavior.  `ret += b;` will work better.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,993 @@
...
+   using std::string::data;
+   using std::string::size;
+};
+
+/* @cond PRIVATE */
+inline key_type make_prefix(eosio::name table_name, eosio::name index_name, uint8_t status = 1) {
+   auto bige_table = swap_endian<uint64_t>(table_name.value);
+   auto bige_index = swap_endian<uint64_t>(index_name.value);
+
+   size_t size_name = sizeof(index_name);
```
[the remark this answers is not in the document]
swatanabe-b1: constexpr is more explicit, but worrying about this level of optimization is a bit absurd.  Constant propagation is an easy optimization.  The compiler really doesn't need help to do it.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,993 @@
...
+
+            void* pk_buffer = actual_data_size > detail::max_stack_buffer_size ? malloc(actual_data_size) : alloca(actual_data_size);
+            internal_use_do_not_use::kv_get_data(config.db_name, 0, (char*)pk_buffer, actual_data_size);
+
+            deserialize_buffer = pk_buffer;
+            deserialize_size = actual_data_size;
+         }
+
+         T val;
+         deserialize(val, deserialize_buffer, deserialize_size);
```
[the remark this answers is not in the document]
swatanabe-b1: I don't understand.  If you want the return value to be constructed in place, you have to rely on NRVO no matter what--if not here, then in the implementation of deserialize.  Also, deserialize is overloaded based on the type of val, so passing T as an explicit template argument would needless complicate things.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,993 @@
...
+      {}
+
+      ~iterator() {
+         if (itr) {
+            internal_use_do_not_use::kv_it_destroy(itr);
+         }
+      }
+
+      iterator& operator=(iterator&& other) {
+         itr = std::exchange(other.itr, itr);
```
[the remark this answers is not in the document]
jeffreyssmith2nd: We need the old iterator to get cleaned up (`kv_it_destroy`) which the destructor does, so my thought was by using exchange the old iterator would get cleaned up properly.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,993 @@
...
+      key_type prefix;
+   };
+
+   class iterator {
+   public:
+      iterator(uint32_t itr, kv_it_stat itr_stat, const index_config& config) : itr{itr}, itr_stat{itr_stat}, config{config} {}
+
+      iterator(iterator&& other) :
+         itr(std::exchange(other.itr, 0)),
+         itr_stat(std::move(other.itr_stat)),
```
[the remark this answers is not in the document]
jeffreyssmith2nd: I used exchange here for similar reasons to my answer above, except by setting it to 0 I know that `kv_it_destroy` will not get called on it based on the destructor.

If after a move, we can be sure `if(itr)` will return false, then I could switch to the default move constructor.

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
swatanabe-b1: It's technically safe, but it's generally better to clean up resources sooner rather than later.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1000 @@
...
+ * @param table_name  - The name of the table
+ * @param db_name     - The type of the EOSIO Key Value database. Defaulted to eosio.kvram
+ * @param ...         - A variadic list of 1 or more indexes to define on the table.
+ */
+#define DEFINE_TABLE(table_class, value_class, table_name, db_name, /*indices*/...)                                    \
+   struct table_class : eosio::kv_table<value_class> {                                                                 \
+      EOSIO_CDT_CREATE_KV_INDICES(value_class, __VA_ARGS__)                                                            \
+                                                                                                                       \
+      table_class(eosio::name contract_name) {                                                                         \
+         init(contract_name, table_name##_n, db_name EOSIO_CDT_LIST_KV_INDICES(__VA_ARGS__));                          \
```
arhag: If you make this
```cpp
init(contract_name, eosio::name(table_name), eosio::name(db_name),  EOSIO_CDT_LIST_KV_INDICES(__VA_ARGS__));   
```
then the user of the macro has the flexibility of providing either a regular string (and not just as a string literal but also as the name of some `const` global variable if desired) or a `""_n` string and it would work either way.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1000 @@
...
+ * @brief Macro to define an index.
+ * @details In the case where the autogenerated index names created by DEFINE_TABLE are not enough, a user can instead
+ * manually define the table and indices. This macro allows users to conveniently define an index without having to specify
+ * the index template type, as those can be large/unwieldy to type out.
+ *
+ * @param index_name    - The index name.
+ * @param member_name   - The name of the member pointer used for the index. This also defines the index's C++ variable name.
+ */
+#define KV_NAMED_INDEX(index_name, member_name)                                                                        \
+   index<EOSIO_CDT_GET_RETURN_T(value_type, member_name)> member_name{index_name ## _n, &value_type::member_name};
```
arhag: Again use `eosio::name(index_name)` rather than `index_name ## _n`.

Also this assumes that the `value_type` type alias is defined in the table struct. For consistency the `DEFINE_TABLE` macro should define this type alias. Also the example documentation to include that and make a note about how it is expected that the `value_type` type alias is properly defined when manually constructing the table struct.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1000 @@
...
+
+inline key_type make_key(const std::string& val, bool case_insensitive=false) {
+   return make_key(val.data(), val.size(), case_insensitive);
+}
+
+inline key_type make_key(const std::string_view& val, bool case_insensitive=false) {
+   return make_key(val.data(), val.size(), case_insensitive);
+}
+
+inline key_type make_key(const char* str, bool case_insensitive=false) {
```
arhag: I didn't bother reviewing any of the `make_key` stuff (or any of the functions that only exist to implement `make_key` functions such as `swap_endian`, `flip_msb`, ` get_msb`, and `make_floating_key`) because from what I understood all this will be replaced by abieos. 

But I also want to confirm that we are going to be removing all the case insensitive stuff, right?

jeffreyssmith2nd: Yes, all case insensitive stuff will be removed.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: The value_type alias is defined in the base table class, so we know it will always be there.

arhag: Oh I see. Still working my way through the code.

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
jeffreyssmith2nd: In that case, would the recommended solution be to call `kv_it_destroy` on the old itr here in the move equals and then not bother with the exchange?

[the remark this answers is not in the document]
arhag: Yes, I think it would be best to call `kv_it_destroy` on the old value of `itr`, set `itr` to `other.itr` and then set `other.itr` to 0 to ensure that it doesn't call `kv_it_destroy` on an already destroyed iterator when the other value's destructor gets called.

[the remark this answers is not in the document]
arhag: I also think it makes sense to set `other.itr_stat` to `status::iterator_end` both here and in the move constructor. That way the other is always in a consistent state (even though the user really shouldn't be doing anything with that move-from iterator except either calling the destructor or assigning to it).

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+      (EOSIO_CDT_CAT(EOSIO_CDT_KV_INDEX_TEST_,                                                                         \
+           EOSIO_CDT_EXPAND(EOSIO_CDT_KV_INDEX_TEST EOSIO_CDT_KV_INDEX ## index_name ()))))(index_name)
+
+#define EOSIO_CDT_KV_INDEX_CONSTRUCT(value_class, index_name)                                                          \
+   EOSIO_CDT_APPLY(EOSIO_CDT_KV_FIX_INDEX_CONSTRUCT,                                                                   \
+      (EOSIO_CDT_CAT(EOSIO_CDT_KV_INDEX_TEST_,                                                                         \
+           EOSIO_CDT_EXPAND(EOSIO_CDT_KV_INDEX_TEST EOSIO_CDT_KV_INDEX ## index_name ()))))(value_class, index_name)
+
+#define EOSIO_CDT_CREATE_KV_INDEX(r, value_class, i, index_name)                                                       \
+   EOSIO_CDT_KV_INDEX_TYPE(index_name)<EOSIO_CDT_GET_RETURN_T(value_class, index_name)> EOSIO_CDT_KV_INDEX_NAME(index_name, i) EOSIO_CDT_KV_INDEX_CONSTRUCT(value_class, index_name);
```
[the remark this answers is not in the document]
jeffreyssmith2nd: As discussed yesterday, the DEFINE_TABLE macro will be removed so all of this code will go away.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,896 @@
...
+      index->tbl = this;
+
+      index->setup();
+      secondary_indices.push_back(index);
+      setup_indices(indices...);
+   }
+
+   template <typename PrimaryIndex, typename... SecondaryIndices>
+   void init(eosio::name contract, eosio::name table, eosio::name db, PrimaryIndex prim_index, SecondaryIndices... indices) {
+      validate_types(prim_index, indices...);
```
swatanabe-b1: A fold expression is somewhat more friendly than recursion, as it reduces the depth of the template instantiation stack in error messages.
You might also consider ``std::enable_if_t<(std::is_base_of_v<kv_index, std::remove_ptr_t<SecondaryIndices>> && ...)>``

When we get C++20, you'll be able to use concepts for this.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,896 @@
...
+      db_name = db.value;
+
+      primary_index = prim_index;
+      primary_index->contract_name = contract_name;
+      primary_index->table_name = table_name;
+      primary_index->tbl = this;
+
+      primary_index->setup();
+
+      setup_indices(indices...);
```
swatanabe-b1: You can use ``(setup_index(indices), ...)`` instead of recursion.

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: Using `enable_if` for both `PrimaryIndex` and `SecondayIndices` gets a bit unwieldy, so I'll probably just stick with a fold expression here.

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
jeffreyssmith2nd: see #850

libraries/eosiolib/contracts/eosio/key_value.hpp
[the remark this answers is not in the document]
jeffreyssmith2nd: see #850

libraries/eosiolib/contracts/eosio/key_value.hpp
jeffreyssmith2nd: see #850

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+   }
+
+   using std::string::data;
+   using std::string::size;
+};
+
+/* @cond PRIVATE */
+inline key_type make_prefix(eosio::name table_name, eosio::name index_name, uint8_t status = 1) {
+   auto bige_table = swap_endian<uint64_t>(table_name.value);
+   auto bige_index = swap_endian<uint64_t>(index_name.value);
```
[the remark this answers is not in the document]
jeffreyssmith2nd: see #850

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,1018 @@
...
+
+/**
+ * The key_type struct is used to store the binary representation of a key.
+ */
+struct key_type : private std::string {
+   key_type() : std::string() {}
+   key_type(const char* c, size_t s) : std::string(c, s) {}
+
+
+   key_type operator+(const key_type& b) const {
```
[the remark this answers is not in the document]
jeffreyssmith2nd: added in #850

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,888 @@
...
+      }
+
+      internal_use_do_not_use::kv_erase(db_name, contract_name.value, tbl_key.data(), tbl_key.size());
+   }
+
+protected:
+   kv_table() = default;
+
+   template <typename I, typename... Indices>
+   void setup_indices(I&& index, Indices&&... indices) {
```
swatanabe-b1: Indices is dead.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,888 @@
...
+   eosio::name table_name;
+   uint64_t db_name;
+
+   kv_index* primary_index;
+   std::vector<kv_index*> secondary_indices;
+
+   constexpr void validate_types() {}
+
+   template <typename Type, typename... Types>
+   constexpr void validate_types(Type&& t, Types&&... ts) {
```
swatanabe-b1: Types is dead.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,888 @@
...
+      idx->table_name = table_name;
+      idx->tbl = this;
+
+      idx->setup();
+      secondary_indices.push_back(idx);
+   }
+
+   template <typename PrimaryIndex, typename... SecondaryIndices>
+   void init(eosio::name contract, eosio::name table, eosio::name db, PrimaryIndex&& prim_index, SecondaryIndices&&... indices) {
+      (validate_types(prim_index, indices), ...);
```
swatanabe-b1: This only validates prim_index, because the other arguments to validate_types are ignored.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,888 @@
...
+      idx->contract_name = contract_name;
+      idx->table_name = table_name;
+      idx->tbl = this;
+
+      idx->setup();
+      secondary_indices.push_back(idx);
+   }
+
+   template <typename PrimaryIndex, typename... SecondaryIndices>
+   void init(eosio::name contract, eosio::name table, eosio::name db, PrimaryIndex&& prim_index, SecondaryIndices&&... indices) {
```
swatanabe-b1: Since you no longer need to allow rvalues, just use `PrimaryIndex&` and `SecondaryIndex&`.  This will also simplify validate_types.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,888 @@
...
+            return ret_val;
+         }
+
+         void* buffer = value_size > detail::max_stack_buffer_size ? malloc(value_size) : alloca(value_size);
+         auto copy_size = internal_use_do_not_use::kv_get_data(config.db_name, 0, (char*)buffer, value_size);
+
+         void* deserialize_buffer = buffer;
+         size_t deserialize_size = copy_size;
+
+         bool is_primary = config.index_name != config.primary_index_name;
```
swatanabe-b1: This variable name doesn't seem right.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -624,10 +617,10 @@ class kv_table {
        * @return An iterator to the object with the lowest key (by this index) in the table.
        */
       iterator begin() const {
-         uint32_t itr = internal_use_do_not_use::kv_it_create(config.db_name, config.contract_name.value, config.prefix.data(), config.prefix.size());
+         uint32_t itr = internal_use_do_not_use::kv_it_create(tbl->db_name, contract_name.value, prefix.data(), prefix.size());
          int32_t itr_stat = internal_use_do_not_use::kv_it_lower_bound(itr, "", 0);
 
-         return {itr, static_cast<typename iterator::status>(itr_stat), config};
+         return {itr, static_cast<typename iterator::status>(itr_stat), reinterpret_cast<const kv_index*>(this)};
```
swatanabe-b1: You don't need a cast for a derived to base conversion.  Using reinterpret_cast is undefined behavior in this case.

jeffreyssmith2nd: When I added this, it failed to find the correct `iterator` constructor otherwise. Removing it now works though.

swatanabe-b1: If you need a cast, the correct cast for this scenario is static_cast.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,889 @@
...
+         iterator_end    = -2, // Iterator is out-of-bounds
+      };
+
+      iterator() = default;
+
+      iterator(uint32_t itr, status itr_stat, const kv_index* index) : itr{itr}, itr_stat{itr_stat}, index{index} {}
+
+      iterator(iterator&& other) :
+         itr(std::exchange(other.itr, 0)),
+         itr_stat(std::move(other.itr_stat))
```
swatanabe-b1: Does not copy index.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,889 @@
...
+            internal_use_do_not_use::kv_it_destroy(itr);
+         }
+      }
+
+      iterator& operator=(iterator&& other) {
+         if (itr) {
+            internal_use_do_not_use::kv_it_destroy(itr);
+         }
+         itr = std::exchange(other.itr, 0);
+         itr_stat = std::move(other.itr_stat);
```
swatanabe-b1: index too

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -0,0 +1,889 @@
...
+         for(auto itr = lower_bound(b), end_itr = lower_bound(e); itr < end_itr; ++itr) {
+            return_values.push_back(itr.value());
+         }
+
+         return return_values;
+      }
+
+      void setup() override {
+         prefix = make_prefix(table_name, index_name);
+      }
```
swatanabe-b1: Shouldn't this be private?

tbfleming:
🎉🎉🎉🎉🎉🎉🎉

PR #808, opened 2020-02-06
Serialize non-variants as 0 index variant for upgrade compatibility

jeffreyssmith2nd:
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -3,6 +3,7 @@
 #include "../../core/eosio/name.hpp"
 #include "../../core/eosio/print.hpp"
 #include "../../core/eosio/utility.hpp"
+#include "../../core/eosio/varint.hpp"
 
 #include <algorithm>
 #include <cctype>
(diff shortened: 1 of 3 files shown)


libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -876,5 +874,46 @@ class kv_table {
...
+      datastream<char*> ds((char*)buffer, size);
+      ds << value;
+   }
+
+   template <typename V>
+   static void deserialize(V& value, void* buffer, size_t size) {
+      unsigned_int idx;
+      datastream<const char*> ds((char*)buffer, size);
+
+      ds >> idx;
```
swatanabe-b1: You should check that idx is 0.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -876,5 +874,46 @@ class kv_table {
...
+   }
+
+   template <typename V>
+   static size_t get_size(V&& value) {
+      auto size = pack_size(std::forward<V>(value));
+      return size + 1;
+   }
+
+   template <typename... Vs>
+   static size_t get_size(std::variant<Vs...>&& value) {
```
swatanabe-b1: This overload will never be chosen, because you're passing an lvalue.

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -876,5 +874,47 @@ class kv_table {
...
+   }
+
+   template <typename... Vs>
+   static void deserialize(std::variant<Vs...>& value, void* buffer, size_t size) {
+      datastream<const char*> ds((char*)buffer, size);
+      ds >> value;
+   }
+
+   template <typename V>
+   static size_t get_size(V&& value) {
```
swatanabe-b1: You should also change this overload to prevent errors.  I don't see any reason for get_size to take an rvalue.

PR #809, opened 2020-02-07
Add support for uppercasing unicode

jeffreyssmith2nd:
## Change Description

Add support for uppercasing unicode strings.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -230,7 +230,8 @@ inline key_type make_key(const char* str, size_t size, bool case_insensitive=fal
    void* data_buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
 
    if (case_insensitive) {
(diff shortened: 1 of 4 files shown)


swatanabe-b1:
Let's back up a bit.  What is the use case for case-insensitive keys?  Correct collation is locale dependent and is much more complex than any variant of toupper is capable of representing. See also https://unicode.org/reports/tr10/

jeffreyssmith2nd:
I was following the recommendation [here](https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database.md#key-transformations) (which does note that it assumes ASCII), but my assumption is simply to provide convenience for developers that may want to do case insensitive comparisons.

If we think generically supporting it ourselves will be too complicated, I'm not opposed to requiring the developers to handle their own case insensitive implementation.

Or, provide the `make_insensitive` function as it was originally and document that it is only compatible with ASCII.

jeffreyssmith2nd:
We've discussed this in person, we are not planning unicode support for toupper at this time.

PR #810, opened 2020-02-11
Feature/action results

larryk85:
## Change Description

Add support for ABI generation and auto code generation for action results.

## API Changes
- [x] API Changes

`eosio actions` can now have a return type other than `void`.

## Documentation Additions
- [x] Documentation Additions

`eosio actions` can now be defined as follows.
```c++
[[eosio::action]]
int action1() {
   ...
   return 42;
}
[[eosio::action]]
std::string action2() {
  std::string default_string_value = "action2 action";
   ...
   if (something)
      return "foo";
   ...
   return default_string_value;
}
```

This will now generate the correct code to call `set_action_return_value` and populate the fields for `action_results` in the ABI.

This PR also introduces a new flag for the compiler `--abi-version` this can be given to generate an ABI that is conformant with currently either v1.1 or v1.2.


tools/toolchain-tester/tests.py
```diff
@@ -97,8 +97,12 @@ def handle_expecteds(self, res: subprocess.CompletedProcess):
...
                 )
 
-        if expected.get("abi"):
-            expected_abi = expected["abi"]
+        if expected.get("abi") or expected.get("abi-file"):
+            if expected.get("abi"):
+                expected_abi = expected["abi"]
+            else:
+                expected_abi_file = open(expected["abi-file"])
+                expected_abi = expected_abi_file.read()
```
jeffreyssmith2nd: This needs either a `expected_abi_file.close()` or

```
with open(expected["abi-file"]) as expected_abi_file:
   expected_abi = ...
```

larryk85: gotcha, thanks for catching that.

tools/include/eosio/abigen.hpp
```diff
@@ -481,10 +496,16 @@ namespace eosio { namespace cdt {
             o["variants"].push_back(variant_to_json( v ));
          }
          o["abi_extensions"]     = ojson::array();
+         if (_abi.version_major == 1 && _abi.version_minor == 2) {
```
tbfleming: minor >= 2

PR #851, opened 2020-03-16
KV Wrapper Singleton

jeffreyssmith2nd:
## Change Description

Add a new class, kv_singleton, for easily storing global objects using the kv intrinsics.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+   private:
+      constexpr static uint64_t db_name = static_cast<uint64_t>(DbName);
+      constexpr static uint64_t singleton_name = static_cast<uint64_t>(SingletonName);
+
+      eosio::name contract_name;
+      key_type key;
+
+      key_type raw_original;
+      bool is_dirty = false;
+      bool is_cached = false;
```
swatanabe-b1: Shouldn't these have the same lifetime as value?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+      key_type make_prefix(uint64_t singleton_name, uint8_t status = 2) {
+         return make_key(std::make_tuple(status, singleton_name));
+      }
+
+      template <typename V>
+      static void deserialize(V& value, void* buffer, size_t size) {
+         datastream<const char*> ds((char*)buffer, size);
+         ds >> value;
+      }
```
swatanabe-b1: Shouldn't this provide the same upgradeable-to-variant behavior as kv_table?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+         is_cached = true;
+      }
+
+      void erase() {
+         internal_use_do_not_use::kv_erase(db_name, contract_name.value, key.data(), key.size());
+         is_cached = false;
+         is_dirty = false;
+      }
+
+      void write_back() {
```
swatanabe-b1: What if I call `write_back` before calling `get` or `set` or after calling `erase`

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+         size_t data_size = get_size(value());
+         void* data_buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
+
+         serialize(value(), data_buffer, data_size);
+
+         if (raw_original.size() != data_size || memcmp(raw_original.data(), data_buffer, data_size) != 0) {
```
swatanabe-b1: What if I call `set` without first calling `get`, and T is an empty struct?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+         __attribute__((eosio_wasm_import))
+         uint32_t kv_get_data(uint64_t db, uint32_t offset, char* data, uint32_t data_size);
+      }
+   }
+
+   template <typename T, eosio::name::raw SingletonName, eosio::name::raw DbName = "eosio.kvram"_n>
+   class kv_singleton {
+   public:
+      kv_singleton(eosio::name contract_name) : contract_name{contract_name} {
```
bytemaster: contract_name = get_self()

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
swatanabe-b1: Also, what if I run something like:
```c++
auto original_value = get();
erase();
write_back();
set(original_value);
write_back();
```

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+         __attribute__((eosio_wasm_import))
+         bool kv_get(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, uint32_t& value_size);
+
+         __attribute__((eosio_wasm_import))
+         uint32_t kv_get_data(uint64_t db, uint32_t offset, char* data, uint32_t data_size);
+      }
+   }
+
+   template <typename T, eosio::name::raw SingletonName, eosio::name::raw DbName = "eosio.kvram"_n>
```
bytemaster: By defining a default parameter here we simplify cases where the type and name are basically the same.

also instead of hard coding "eosin.kvram"_n we should use eosin::kv_ram constant 
```suggestion
   template <typename T, eosio::name::raw SingletonName="anonymous"_n, eosio::name::raw DbName = eosio::kv_ram>
```

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
bytemaster:
```suggestion
      void store() {
```

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
swatanabe-b1: You have to give the singleton a name.  Otherwise you can only have one singleton (one **total** not per-type) in your whole contract.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+   class kv_singleton {
+   public:
+      kv_singleton(eosio::name contract_name) : contract_name{contract_name} {
+         key = make_prefix(singleton_name);
+      }
+
+      ~kv_singleton() {
+         if (is_dirty) {
+            write_back();
```
bytemaster:
```suggestion
            store();
```

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+         __attribute__((eosio_wasm_import))
+         bool kv_get(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, uint32_t& value_size);
+
+         __attribute__((eosio_wasm_import))
+         uint32_t kv_get_data(uint64_t db, uint32_t offset, char* data, uint32_t data_size);
+      }
+   }
+
+   template <typename T, eosio::name::raw SingletonName, eosio::name::raw DbName = "eosio.kvram"_n>
+   class kv_singleton {
```
bytemaster: this isn't really a "kv_singleton". "kv" is an implementation detail, let's just call it. single<T>,

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,129 @@
...
+         key = make_prefix(singleton_name);
+      }
+
+      ~kv_singleton() {
+         if (is_dirty) {
+            write_back();
+         }
+      }
+
+      const T& get() {
```
bytemaster:
```suggestion
      const T& get() const {
```

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: We already have a `singleton` class that was implemented using multi-index, which is why I stuck `kv` on the front.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
swatanabe-b1: get_self is a member of eosio::contract and is not visible here.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
bytemaster: These fields need the same lifetime as the value... otherwise I can create N copies of single<T> and each will have a different dirty/cached bit but all return the same value.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: I'm open to a different idea for a name that doesn't conflict with that though.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
bytemaster: You do give it a name if you want something other than the default... oh the name needs to include info from T.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
bytemaster: I thought it was something that could be called from the C API intrinsic

jeffreyssmith2nd: I don't believe so. It's just a member of eosio::contract that gets set when the contract is instantiated.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: Added a `is_dirty` check that should handle this case.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: In the first scenario, now that it has the upgrade-to-variant behaviour an empty struct should serialize to something with a size 1, so it will call the intrinsic.

I tested the second scenario with my latest changes, and it works as expected.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,164 @@
...
+         auto& ste = get_state();
+         if (ste.is_dirty) {
+            size_t data_size = get_size(ste.value);
+            void* data_buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
+
+            serialize(ste.value, data_buffer, data_size);
+
+            if (ste.raw_original.size() != data_size || memcmp(ste.raw_original.data(), data_buffer, data_size) != 0) {
+               internal_use_do_not_use::kv_set(db_name, contract_name.value, key.data(), key.size(), (const char*)data_buffer, data_size);
+            }
```
swatanabe-b1: `ste.is_dirty = false;`?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+      void store() {
+         auto& ste = get_state();
+         if (ste.is_dirty) {
+            size_t data_size = get_size(ste.value);
+            void* data_buffer = data_size > detail::max_stack_buffer_size ? malloc(data_size) : alloca(data_size);
+
+            serialize(ste.value, data_buffer, data_size);
+
+            if (ste.raw_original.size() != data_size || memcmp(ste.raw_original.data(), data_buffer, data_size) != 0) {
```
swatanabe-b1: Is it a problem that bypasses access control and allows any contract to call set as long as it sets it to its original value?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+   private:
+      constexpr static uint64_t db_name = static_cast<uint64_t>(DbName);
+      constexpr static uint64_t singleton_name = static_cast<uint64_t>(SingletonName);
+
+      eosio::name contract_name;
+      key_type key;
+
+      key_type make_prefix(uint64_t singleton_name, uint8_t status = 2) {
```
swatanabe-b1: Do you need these arguments?  You're never using different values.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,164 @@
...
+      static size_t get_size(const V& value) {
+         auto size = pack_size(value);
+         return size + 1;
+      }
+
+      template <typename... Vs>
+      static size_t get_size(const std::variant<Vs...>& value) {
+         auto size = pack_size(value);
+         return size;
+      }
```
swatanabe-b1: This duplicates code in key_value.hpp

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
```diff
@@ -0,0 +1,164 @@
...
+
+      const T& get() {
+         auto& ste = get_state();
+         if (!ste.is_cached) {
+            uint32_t copy_size;
+            uint32_t value_size;
+
+            auto success = internal_use_do_not_use::kv_get(db_name, contract_name.value, key.data(), key.size(), value_size);
+
+            eosio::check(success, "tried to get a singleton that does not exist");
```
swatanabe-b1: Do we need a way to check whether the singleton exists?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
swatanabe-b1: A stranger case is setting it to a new value, and then setting it back to the old value.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
swatanabe-b1: Really? I think there's still a problem with the second case.  I thought that raw_original would still hold the serialized representation of original_value, since raw_original is only changed by get.  Thus, the final store will be a no-op.  Am I missing something?

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: What do you mean by bypasses access control?

swatanabe-b1: I mean the restriction that a contract can only write to its own kv store.

jeffreyssmith2nd: Doesn't the `contract_name.value` in `internal_use_do_not_use::kv_set(db_name, contract_name.value, key.data(), key.size(), (const char*)data_buffer, data_size);` handle that restriction? I thought the intrinsic would enforce that the contract passed in is the same as the contract that its running in.

swatanabe-b1: It does--but only if kv_set is actually called.  To be clear, this doesn't allow the contract to cause problems for another contract.  It just means that the caching behavior is observable and code that is probably incorrect will sometimes "work."

jeffreyssmith2nd: Ah I understand now. I've considered removing this specific check, and just always call `kv_set` if `is_dirty` to simplify the code.

libraries/eosiolib/contracts/eosio/key_value_singleton.hpp
jeffreyssmith2nd: Yeah I'm planning on moving this to the `key_value.hpp` file once all the other comments are resolved (I'm not sure GH would track that move correctly).

jeffreyssmith2nd:
Closed in favor of #924

PR #864, opened 2020-04-12
Fix/non type template args

larryk85:
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


jeffreyssmith2nd:
Looks like this caused a build failure on 16.04

jeffreyssmith2nd:
> Looks like this caused a build failure on 16.04

More specifically, it introduces a dependency on `std::variant` which means we need a compiler/stdlib version increase (specifically for the `tools` project).

PR #873, opened 2020-04-22
Update kv_table example

dixia:
Update kv_table Example including:
* Use up-to-date header
* Fix minor C++ syntax
* Update the member variable reference used for index

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


dixia:
To be fix the `myrecord` issue after confirmation

dixia:
The updated code example still cannot be compiled.Tthe error log here:  https://gist.github.com/dixia/01d405c76ff081915eb34f8c1977f817

jeffreyssmith2nd:
The `full_name` function needs to be const.

`auto full_name() const ...`

PR #880, opened 2020-05-13
Remove scripts that are no longer needed

jeffreyssmith2nd:
## Change Description

These scripts seem to not be used anywhere and to make sure everyone that builds from source builds the recommended way, these should be removed.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


scottarnette:
Looks like the readme points to using cmake/make as the official build method. No complaints removing these scripts!

PR #917, opened 2020-06-26
Don't default to merging data sections, it can blow out nodeos limits

jeffreyssmith2nd:
## Change Description

In LLVM9, special handling was added for `rodata` when creating the data segments. Before that, it was not handled and therefore was always not merged. This restores that behaviour to prevent having segments that are too large.

An alternative would be to remove the handling inside lld, but this feels like the less intrusive solution and clearly communicates the behaviour we want.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -10,7 +10,7 @@ steps:
       IMAGE_TAG: "amazonlinux-2"
     agents:
       queue: "automation-eks-eos-builder-fleet"
-    timeout: ${TIMEOUT:-30}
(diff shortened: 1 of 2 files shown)


jeffreyssmith2nd:
Based on my test with eosio.system from eosio.contracts, and the 2 example apps here in CDT, this produces nearly identical data sections.

Which I think should make sense, right? The strings should be read-only data and since that wasn't being handled before, they weren't getting merged anyways.

PR #918, opened 2020-06-26
Fix test_chain::get_history

swatanabe-b1:
Passing a const argument to build_history_result causes it always to choose the error overload.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -10,7 +10,7 @@ steps:
       IMAGE_TAG: "amazonlinux-2"
     agents:
       queue: "automation-eks-eos-builder-fleet"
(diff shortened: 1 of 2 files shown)


jeffreyssmith2nd:
@swatanabe-b1 The CI timeout failure seems to have started sometime recently. It was defaulting to no timeout at all, and then something changed.

swatanabe-b1:
The timeout has been set to 10 minutes for a while, but was ignored for some reason.
```
42833e2544 (Scott Arnette    2019-08-23 20:02:55 -0400  13)     timeout: ${TIMEOUT:-10}
```
It seems like it was fixed, and we just need to increase the timeout.

PR #923, opened 2020-07-14
Update toolchain-tester to allow for changes in minor version of Python

jeffreyssmith2nd:
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


jeffreyssmith2nd:
This change was made since the latest Anka Mac builders are now pulling Python 3.8. Rather than having to bump the minor every time we move to a new Python package, we should just depend on Python 3.

jeffreyssmith2nd:
This PR needs more changes, closing in favor of a future PR that is correct.

PR #930, opened 2020-08-03
Add ABI Generation for KV Tables

jeffreyssmith2nd:
## Change Description

This change introduces abi generation for KV tables as well as a host of refactorings to abigen and codegen in general.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


jeffreyssmith2nd:
**Note**: This is currently targeted at the `kv-wrapper-mega-merge` branch so the diff looks correct. However, it should really just be merged to develop once all of the underlying KV work is merged.

tools/include/eosio/abigen.hpp
```diff
@@ -530,5 +626,143 @@ namespace eosio { namespace cdt {
...
+               auto type = base.getType();
+               if (type.getAsString().find("eosio::kv_table<") != std::string::npos) {
+                  return true;
+               }
+            }
+
+            return false;
+         }
+
+         std::string get_value_from_decltype(const clang::DecltypeType* decl) {
```
larryk85, line 641: This isn't really getting the value from a decltype.  It is getting the std::string representation of the deduced type.

tools/include/eosio/abigen.hpp
```diff
@@ -530,5 +626,143 @@ namespace eosio { namespace cdt {
...
+            if (const auto ref_type = dyn_cast<clang::LValueReferenceType>(decl->desugar())) {
```
larryk85, line 642: Do we only care about the case where the argument is an lvalue?  It is perfectly legal to call `decltype` on literals and rvalues and even functions.  

I am fine with the specificity of this function if:
  * it is named in a way that shows that it is not a general use function for getting the type from a decltype
  * we control the code on the CDT side that would hit this code and it wouldn't make much sense to ever use any of the above mentioned arguments (limit the chance of someone coming in later and causing ABI generation internal errors)

tools/include/eosio/abigen.hpp
```diff
@@ -530,5 +626,143 @@ namespace eosio { namespace cdt {
          std::set<abi_table>                   ctables;
          std::map<std::string, std::string>    rcs;
          std::set<const clang::Type*>          evaluated;
+
+         bool is_kv_table(const clang::CXXRecordDecl* decl) {
```
larryk85, line 630: This function and `get_value_from_decltype` and `internal_types` are generic enough that they should live in `gen.hpp`.

tools/include/eosio/abigen.hpp
```diff
@@ -530,5 +626,143 @@ namespace eosio { namespace cdt {
...
+               } else {
+                  // Handle non-records
+                  return get_type(pt);
+               }
+            }
+            CDT_INTERNAL_ERROR("Error generating type from kv index definition");
+            return "";
+         }
+
+         const std::set<std::string> internal_types {
```
larryk85, line 673: These types should be namespace qualified.  It seems like this could cause issues with user defined types being flagged incorrectly.

tools/include/eosio/codegen.hpp
```diff
@@ -272,23 +175,13 @@ namespace eosio { namespace cdt {
...
-                  ss << ");\n";
-               };
-               if (return_ty != "void") {
-                  ss << "const auto& result = ";
-               }
-               call_action();
-               if (return_ty != "void") {
-                  ss << "const auto& packed_result = eosio::pack(result);\n";
-                  ss << "set_action_return_value((void*)packed_result.data(), packed_result.size());\n";
+               ss << decl->getParent()->getQualifiedNameAsString() << "{eosio::name{r},eosio::name{c},ds}." << decl->getNameAsString() << "(";
```
larryk85, line 178: We shouldn't have gotten rid of `call_action` and the rest as it's used for action results.

tools/include/eosio/gen.hpp
```diff
@@ -300,8 +304,8 @@ struct generation_utils {
                   return true;
                } else {
                   for ( auto name : names )
-                     if ( const auto* decl = rt->getDecl() ) {
-                        return decl->getName().str() == name;
+                     if ( rt->getDecl()->getName().str() == name ) {
```
larryk85, line 307: I don't understand why this changed.  `rt->getDecl()` can't guarantee that the decl pointer is non-null.

tools/include/eosio/gen.hpp
```diff
@@ -446,6 +448,8 @@ struct generation_utils {
          {"unsigned_int", "varuint32"},
          {"signed_int",   "varint32"},
 
+         {"basic_string<char>", "string"},
```
larryk85, line 451: Should we care about any other `basic_string<T>` types where `T` is 8 bits?

tools/include/eosio/gen.hpp
```diff
@@ -580,6 +588,10 @@ struct generation_utils {
       return builtins.count(_translate_type(t)) >= 1;
    }
 
+   inline bool is_reserved_type( const std::string& t ) {
```
larryk85, line 591: This doesn't have the same calling style as the rest of the `is_*_type` functions.  Is that by design?

tools/include/eosio/gen.hpp
```diff
@@ -615,5 +627,25 @@ struct generation_utils {
       if (get_base_type_name(t).find("<") != std::string::npos) return false;
       return get_base_type_name(t).compare(get_type_alias_string(t)) != 0;
    }
+
+   inline bool is_eosio_non_unique(const clang::QualType& t) {
+      auto str_name = t.getAsString();
+      auto nu_re = std::regex("eosio::non_unique<[a-zA-Z]+[a-zA-Z0-9].*>");
```
larryk85, line 633: We don't really care what the template name is as this simply a predicate for the template name.  There's no need to build up the regex for these. 

```c++
constexpr std::string_view test_string = "eosio::non_unique<";
return t.getAsString().substr(0, test_string.size()) == test_string;
```

tools/init/eosio-init.cpp
```diff
@@ -23,6 +23,7 @@
 #include <map>
 #include <chrono>
 #include <ctime>
+#include <regex>
```
larryk85, line 26: No need to include this.

tools/include/eosio/codegen.hpp
jeffreyssmith2nd: Good catch, when I brought the `spec_stuff` branch in, it didn't have action results and I missed adding them.

tools/include/eosio/gen.hpp
jeffreyssmith2nd: Hmm, I don't remember touching this so maybe it was a bad merge. Regardless, I'll revert to the original form.

tools/include/eosio/gen.hpp
jeffreyssmith2nd: Not intentionally, you're right it should either take a QualType or have a different naming format.

tools/include/eosio/abigen.hpp
jeffreyssmith2nd: We (kind of) control the code that hits this because its only called currently in the case where we hit an index that has been defined using the KV_INDEX macro, which will always be an lvalue ref. We could either:

1. Make this more generic to handle many different types of `decltype` calls
2. Rename this function to make it clear that its being used in the specific case mentioned above.

larryk85: I'm fine with just renaming the function for now.  If we need it to be more generic in the future then we can do that then.

tools/include/eosio/abigen.hpp
```diff
@@ -237,12 +250,86 @@ namespace eosio { namespace cdt {
...
+                                 }
+                              }
+                           }
+                        }
+                     }
+                  }
+               }
+            }
+            return index_name;
+         };
```
jeffreyssmith2nd, line 306: If there's a less obnoxious way to do this, I'm all ears.

PR #931, opened 2020-08-03
 Fix problem with action parameters with a template type of bool

### LINKED ISSUE #862: Allow `optional<bool>` as action parameter
Getting an error whenever I enter an `optional<bool>` however it does work for other EOSIO types such as `optional<name>`

This seems like a valid syntax in C++
https://www.boost.org/doc/libs/1_56_0/libs/optional/doc/html/boost_optional/tutorial/a_note_about_optional_bool_.html

## Code example

```c++
#include <eosio/eosio.hpp>
#include <optional>

using namespace eosio;
using namespace std;

class [[eosio::contract]] hello : public contract {
  public:
      using contract::contract;

      [[eosio::action]]
      void hi( optional<bool> option ) {
         if ( option == true ) print( "print me" );
      }
};
```

## ❌ error message

```bash
/var/folders/sc/k5bwq66j0g16dn721lwhvf540000gn/T//hello.cpp:32:15: error: use of undeclared identifier '_Bool'
std::optional<_Bool> arg1; ds >> arg1;
              ^
/var/folders/sc/k5bwq66j0g16dn721lwhvf540000gn/T//hello.cpp:32:22: error: C++ requires a type specifier for all
      declarations
std::optional<_Bool> arg1; ds >> arg1;
                     ^
2 errors generated.
```

## expected ABI

```json
{
    "name": "hi",
    "base": "",
    "fields": [
        {
            "name": "option",
            "type": "bool?"
        }
    ]
}
```

## Inline `optional<bool>` does work ✅ 

```c++
[[eosio::action]]
void hi( ) {
  optional<bool> option = true;
  if ( option == true ) print( "print me" );
}
```

jeffreyssmith2nd:
## Change Description

Handle action parameters that have a template type that translates to `_Bool` during codegen.

Fixes #862.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/codegen.hpp
+++ b/tools/include/eosio/codegen.hpp
@@ -185,6 +185,14 @@ namespace eosio { namespace cdt {
             return true;
          }
 
+         void normalize_type_name(std::string& type_name) {
+            // Currently _Bool is the only one of these oddities but others could be handled here if they arise.
+            auto pos = type_name.find("_Bool");
+            if (pos != std::string::npos) {
+               type_name.replace(pos, 5, "bool");
+            }
+         }
+
          template <size_t N>
          void emitError(CompilerInstance& inst, SourceLocation loc, const char (&err)[N]) {
             FullSourceLoc full(loc, inst.getSourceManager());
@@ -268,7 +276,7 @@ namespace eosio { namespace cdt {
                   qt.removeLocalVolatile();
                   qt.removeLocalRestrict();
                   std::string tn = clang::TypeName::getFullyQualifiedName(qt, *(cg.ast_context), policy);
-                  tn = tn == "_Bool" ? "bool" : tn; // TODO look out for more of these oddities
+                  normalize_type_name(tn);
                   ss << tn << " arg" << i << "; ds >> arg" << i << ";\n";
                   i++;
                }


tools/include/eosio/codegen.hpp
```diff
@@ -185,6 +185,14 @@ namespace eosio { namespace cdt {
             return true;
          }
 
+         void normalize_type_name(std::string& type_name) {
+            // Currently _Bool is the only one of these oddities but others could be handled here if they arise.
+            auto pos = type_name.find("_Bool");
+            if (pos != std::string::npos) {
+               type_name.replace(pos, 5, "bool");
```
dimas1185, line 192: Is this 5 is actually `std::string("_Bool").size()` ?
Also, I assume you expect that whole string should be equal to "_Bool" and not somewhere in the middle.
If so implementation could be like that:
`static const std::string _Bool = "_Bool";
if (type_name == _Bool){
   type_name.replace(0, _Bool.size(), "bool");`
also I'm not sure why are you using `replace`, is it faster comparing to operator assignment? I assume it should have same performance

jeffreyssmith2nd: Replace is needed because _Bool is **not** the whole string. This change was specifically to handle cases like

`std::optional<_Bool>` which should be converted to `std::optional<bool>`

tbfleming: Expanding on the question: what will this do with `std::optional<My_Bool>`?

tbfleming: Or, `std::optional<MyNS::_Bool>`?

C++ types don't meet the requirements of a regular grammar. i.e. neither simple string ops or even regular expressions can disambiguate them.

dimas1185: @jeffreyssmith2nd got it. do you mind to replace hardcoded 5 with `_Bool.size()` ?

jeffreyssmith2nd: > C++ types don't meet the requirements of a regular grammar. i.e. neither simple string ops or even regular expressions can disambiguate them.

@tbfleming Certainly, and that's challenge of some of the string based stuff we do in abi/codegen. In this case, I could add a regex to handle some common scenarions (right after a `<`, comma, or start of a line, etc) which I think would be worth doing to handle cases like what you showed, but I'm not sure there's a solution that handles everything perfectly.

tbfleming: As long as it's string based, there's a tradeoff between:
1. Not recognizing some cases, e.g. `binary_extension<optional<vector<bool>>>`
1. Improperly transforming cases which shouldn't match

Saying abigen doesn't recognize specific cases is easier than saying it scrambles certain user-defined types.

tbfleming: Especially since working contract code could currently exist which this breaks.

swatanabe-b1: `_Bool` only appears because `policy.Bool` is `false`.

jeffreyssmith2nd: Good find, that's probably the right way to fix this.

PR #942, opened 2020-08-06
add payer support for kv table

huangminghuang:
## Change Description
Allow application developers flexibility to choose who pays for resources. 

## API Changes
- [x] API Changes

`kv_table::put()` has an additional optional parameter `payer` 

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -888,10 +888,6 @@ public:
       kv_table_base::put(&value, &old_value, &get_size_fun, &deserialize_fun, &serialize_fun, payer);
    }
 
-   void put(const T& value) {
-      put(value, kv_table_base::contract_name);
-   }
-
    /* @cond PRIVATE */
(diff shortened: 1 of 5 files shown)


libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -880,11 +881,15 @@ class kv_table : kv_detail::kv_table_base {
...
-   void put(const T& value) {
-      uint32_t value_size;
+   void put(const T& value, eosio::name payer) {
       T old_value;
-      kv_table_base::put(&value, &old_value, &get_size_fun, &deserialize_fun, &serialize_fun);
+      kv_table_base::put(&value, &old_value, &get_size_fun, &deserialize_fun, &serialize_fun, payer);
+   }
+
+   void put(const T& value) {
+      put(value, kv_table_base::contract_name);
```
jeffreyssmith2nd, line 892: My preference would be to have KV mirror multi-index and not provide a default like this and instead always require the payer to be specified. The contract can use `get_self()` to pay for the resources if they choose.

PR #948, opened 2020-08-10
Better fix for booleans as template arguments

jeffreyssmith2nd:
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


swatanabe-b1:
`_detail` is reserved in the global namespace.  `_Bool` is also reserved, but using it is unavoidable for this test.  See [lex.name]

PR #949, opened 2020-08-10
Combine unit test for string_tests 

tony-b1:
Fix unit tests regarding splitting the string_tests due to segfault.

## Change Description
This requires the EOSIO.cdt native to have a larger output stream buffer.
This only affects the native eoslib.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/native/native/eosio/crt.hpp
```diff
@@ -8,7 +8,7 @@ namespace eosio { namespace cdt {
       none
    };
    struct output_stream {
-      char output[1024*2];
+      char output[1024*4];
```
revl, line 11: I understand that this fixes the bus error in the unit test. However, increasing the buffer size fixes a symptom rather than the underlying problem. The current `output_stream` implementation is a disaster waiting to happen (again). We need to either fix the implementation or at least put safeguards to check for buffer overruns.

jeffreyssmith2nd:
There's a better long term solution to the problem in the pipeline, I'm going to close this.

PR #957, opened 2020-08-21
Final CDT key value cleanup

allenhan2:
Currently, the key value API in EOSIO.CDT has the prefix 'kv_' used almost everywhere.  This is problematic in a couple of ways, one is that the user has to type 'kv_' a lot and it is a bit reminiscent of C style "namespacing".  So, we should simply add the namespace 'kv' over these data types.  These names will also need to be updated in tests, abigen, and docs.

It also, appears that 'kv_singleton' either made its way back in or it fully wasn't removed in any of the previous PRs.  This data type can be massive foot gun for developers and it is difficult to concisely express the semantics of it.  The remedy is to just remove the data type 'kv_singleton' and any usages from the tests.
## API Changes
- [x] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/docs/06_how-to-guides/06_key_value/01_key_value_examples.md
+++ b/docs/06_how-to-guides/06_key_value/01_key_value_examples.md
@@ -1,4 +1,4 @@
-### kv_table Example:
+### kv::table Example:
 ```cpp
 #include <eosio/eosio.hpp>
 using namespace eosio;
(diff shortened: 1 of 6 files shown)


jeffreyssmith2nd:
Verified tests, everything is passing

PR #961, opened 2020-08-28
[docs] kv api example contract

iamveritas:
adding kv api example contract


examples/kv_addr_book/include/kv_addr_book.hpp
```diff
@@ -0,0 +1,153 @@
...
+   //    it ensures the uniques of the values combined (including non-unique ones)
+   // 3. the rest of the properties are the ones wanted to be indexed non-uniquely
+   index<non_unique<eosio::name, string>> first_name_idx {
+      "firstname"_n, &person::first_name};
+   index<non_unique<eosio::name, string>> last_name_idx {
+      "lastname"_n, &person::last_name};
+   index<non_unique<eosio::name, string>> personal_id_idx {
+      "persid"_n, &person::personal_id};
+   index<non_unique<eosio::name, string, string, string, string>> address_idx {
+      "address"_n, &person::street_city_state_cntry};
```
jeffreyssmith2nd, line 85: I'd like to see the KV_NAMED_INDEX macro demo'd for at least one of these indices. I imagine most people will use the macro and it would make the example look a little less intimidating by removing a lot of code.

examples/kv_addr_book/src/kv_addr_book.cpp
```diff
@@ -0,0 +1,157 @@
...
+      print_person(person_found);
+      // return found person from action
+      return person_found;
+   }
+   else {
+      eosio::print_f("Person not found in addressbook.");
+      // return empty person from action
+      return person{};
+   }
+}
```
jeffreyssmith2nd, line 59: I'd like to see one of these use `get()` to demonstrate more of the API.

halsaphi:
A comment to explain the magic numbers in the getters would be useful:

i.e.

string get_last_name() const {
      return get<1>(this->last_name);
 }

halsaphi:
+ comment on 'using' and 'action_wrapper' - explain why you use aliasing and the benefit of action_wrapper.

+ is it possible to highlight differences in using kv api vs db api?

iamveritas:
> * comment on 'using' and 'action_wrapper' - explain why you use aliasing and the benefit of action_wrapper.
> * is it possible to highlight differences in using kv api vs db api?

these are good thoughts but they are out of scope of this code example
for first one we have documentation to cover it, if user doesn't know what an action wrapper is s/he will look it up and find it.
for second it will be explained in the explainer page

PR #962, opened 2020-09-01
[docs] add kv address example contract

iamveritas:
add kv address example contract


examples/kv_addr_book/include/kv_addr_book.hpp
```diff
@@ -0,0 +1,163 @@
...
+            .first_name = {account_name, first_name},
+            .last_name = {account_name, last_name},
+            .street_city_state_cntry = {account_name, street, city, state, country},
+            .personal_id = {account_name, personal_id},
+            .country_personal_id = {country, personal_id}
+         };
+      }
+};
+
+struct address_table : kv_table<person> {
```
jeffreyssmith2nd, line 69: This should be moved inside your contract class and annotated with `[[eosio::table]]` just like multi_index so that abi generation will find it.

jeffreyssmith2nd (approved):

PR #965, opened 2020-09-03
Fix abi names in KV

jeffreyssmith2nd:
Fixed small typos

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #966, opened 2020-09-03
Add integration tests to CI

jeffreyssmith2nd:
## Change Description

Depends on #965 to pass integration tests

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -37,10 +37,10 @@ steps:
 
   - label: ":ubuntu: Ubuntu 18.04 - Build"
     command:
-      - "./.cicd/build.sh"
+      - "./.cicd/build-integration.sh"
       - "tar -pczf build.tar.gz build && buildkite-agent artifact upload build.tar.gz"
     env:
-      IMAGE_TAG: "ubuntu-18.04"
+      IMAGE_TAG: "eosio/ci-contracts-builder:base-ubuntu-18.04-develop"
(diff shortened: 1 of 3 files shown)


.cicd/pipeline.yml
```diff
@@ -292,7 +292,7 @@ steps:
     agents:
       queue: "automation-eks-eos-tester-fleet"
     timeout: ${TIMEOUT:-10}
-    skip: ${SKIP_UBUNTU_18}${SKIP_UNIT_TESTS}
+    skip: ${SKIP_UBUNTU_18}${SKIP_INTEGRATION_TESTS}
```
linhuang-blockone, line 295: Why is integration test to be skipped?

linhuang-blockone (approved):

.cicd/pipeline.yml
jeffreyssmith2nd: This flag gives you the option to skip the tests by setting an ENV var on a specific build. This is really useful when debugging a different part of the CI pipeline.

linhuang-blockone: Thanks for your explanation! Didn't realize it is a flag.

scottarnette (approved):
Noted the comment that this will fail until a different PR is merged. If that change isn't merging soon, may just want to set the SKIP flag temporarily.
🍕

PR #968, opened 2020-09-08
Fix some issues when building EOSIO contracts

swatanabe-b1:
## Change Description

- Fix dependency file generation (-MD/-MMD) to hold the original source and not the generated temporary file.
- Make temporary filenames unique to avoid races in parallel builds.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/cc/eosio-cpp.cpp.in
+++ b/tools/cc/eosio-cpp.cpp.in
@@ -180,9 +180,6 @@ int main(int argc, const char **argv) {
    try {
       for (auto input : opts.inputs) {
          std::vector<std::string> new_opts = opts.comp_options;
-         SmallString<64> res;
-         llvm::sys::path::system_temp_directory(true, res);
-         std::string tmp_file = std::string(res.c_str())+"/"+llvm::sys::path::filename(input).str();
          std::string output;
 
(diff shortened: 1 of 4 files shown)


arhag (changes requested):
There are two changes I have noticed in the behavior of `eosio-cpp` with these changes. First, the default name chosen for the `.wasm` and `.abi` output files (if not explicitly specified) changes with this patch to something unrelated to the original source file name. Second, even if the compilation is successful it returns a -1 exit code.

That latter regression (at least) seems to be causing all the toolchain tests to fail. I'm not confident if there are other regressions introduced by this patch that may be a source of other problems (whether caught by existing tests or not).

It would be ideal if @larryk85 could also take a look at these changes since he is more familiar with this part of the code.

In addition, @larryk85 you should be aware that this patch fixes bugs (ignoring the regressions/bugs the current version of the patch introduces) with the CDT compiler. So this patch (or rather something like it) should probably be ported not only to the `develop` branch but also any supported stable release branches.

swatanabe-b1:
Both issues are the same problem.

arhag (approved):

PR #973, opened 2020-09-25
[docs]  clean up, get rid of not needed `eosio::` where possible

iamveritas:
we are making use of the `eosio` namespace so all the `eosio::` not necessary are just clobbering the code.

### DIFF
--- a/examples/kv_addr_book/include/kv_addr_book.hpp
+++ b/examples/kv_addr_book/include/kv_addr_book.hpp
@@ -2,44 +2,44 @@
 using namespace std;
 using namespace eosio;
 
-// this structure defines the data stored in the kv_table
+// this structure defines the data stored in the kv::table
 struct person {
-   eosio::name account_name;
-   eosio::non_unique<eosio::name, string> first_name;
-   eosio::non_unique<eosio::name, string> last_name;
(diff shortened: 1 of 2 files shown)


heifner:
In my opinion this PR is a step in the wrong direction. Documentation I believe should be as descriptive as possible. I would think new users would appreciate the namespace indication to make it clear where a particular component comes from. I've always felt good example code should use explicit namespace for each type or a very close by `using` statement.

iamveritas:
> In my opinion this PR is a step in the wrong direction. Documentation I believe should be as descriptive as possible. I would think new users would appreciate the namespace indication to make it clear where a particular component comes from. I've always felt good example code should use explicit namespace for each type or a very close by `using` statement.

iamveritas:
> In my opinion this PR is a step in the wrong direction. Documentation I believe should be as descriptive as possible. I would think new users would appreciate the namespace indication to make it clear where a particular component comes from. I've always felt good example code should use explicit namespace for each type or a very close by `using` statement.

tbfleming:
> another weird thing which I just notice is that I can not get rid of all the eosio::, the ones in the eosio attributes, e.g. [[eosio::contract]], can not lose it.

Correct. C++ attributes don't pay attention to scope.

PR #984, opened 2020-10-22
workround of abi generator bug for key types

vzqzhang:
## Change Description

According to Steven, ABI generator has a bug to generate types of indices in KV tables: https://github.com/EOSIO/eosio.cdt/blob/develop/tools/include/eosio/abigen.hpp#L311-L323. A workaround is to use unqualified names,  like name  instead of eosio::name.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/examples/kv_addr_book/include/kv_addr_book.hpp
+++ b/examples/kv_addr_book/include/kv_addr_book.hpp
@@ -1,5 +1,8 @@
 #include <eosio/eosio.hpp>
 
+using namespace eosio;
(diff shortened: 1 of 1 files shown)


arhag (commented):
The changes here are fine. So it looks good to me in terms of getting this PR in.

I assume there is a separate task for someone to actually fix the root cause of this bug though within CDT.

nksanthosh (approved):

PR #987, opened 2020-10-27
add KV Table iterator bool operator

vzqzhang:
## Change Description

Added bool operator to KV table iterator so that developers can easily check if the iterator exists by using an `if (itr) {}` statement.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -462,6 +462,11 @@ namespace internal {
          return eosio::key_type{(char*)buffer, actual_value_size};
       }
 
+      bool valid() const {
+         eosio::check(itr_stat != status::iterator_erased, "Iterator was erased");
+         return true;
+      }
+
    protected:
       uint32_t itr;
       status itr_stat;
@@ -589,7 +594,7 @@ private:
       }
 
       explicit operator bool() {
-         return base_iterator::compare(this) != 0;
+         return this.valid();
       }
    };
 
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -591,7 +591,7 @@ private:
       }
 
       explicit operator bool() const {
-         return this.valid();
+         return this->valid();
       }
    };
 
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -462,10 +462,7 @@ namespace internal {
          return eosio::key_type{(char*)buffer, actual_value_size};
       }
 
-      bool valid() const {
-         eosio::check(itr_stat != status::iterator_erased, "Iterator was erased");
-         return true;
-      }
+      bool valid() const { return itr_stat != status::iterator_erased; }
 
    protected:
       uint32_t itr;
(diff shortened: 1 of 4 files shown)


libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -587,6 +592,10 @@ class table : internal::table_base {
       bool operator>=(const iterator& b) const {
          return base_iterator::compare(b) >= 0;
       }
+
+      explicit operator bool() {
```
larryk85, line 596: this should be `const`

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -462,6 +462,11 @@ namespace internal {
          return eosio::key_type{(char*)buffer, actual_value_size};
       }
 
+      bool valid() const {
+         eosio::check(itr_stat != status::iterator_erased, "Iterator was erased");
```
larryk85, line 466: This should just return true or false.  I wouldn't expect it to throw an assertion.  That ultimately means that you can't do something like `if(itr)` as it will either return true or assert, if that is the case then using `itr.valid()` and doing away with the `if` statement entirely would get you the same invariants.  Nonetheless, this is not what we want to occur, we want people to be able to query whether the iterator is valid without it throwing an error.

Also, this should be testing if the iterator is erased or it is and end iterator (there are 3 statuses for iterators).

larryk85 (changes requested):

libraries/eosiolib/contracts/eosio/key_value.hpp
vzqzhang: Updated

libraries/eosiolib/contracts/eosio/key_value.hpp
vzqzhang: Updated

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -462,6 +462,8 @@ namespace internal {
          return eosio::key_type{(char*)buffer, actual_value_size};
       }
 
+      bool valid() const { return itr_stat != status::iterator_erased; }
```
larryk85, line 465: Shouldn't this be `itr_stat == status::iterator_ok`?

larryk85 (changes requested):

libraries/eosiolib/contracts/eosio/key_value.hpp
vzqzhang: Should the iterator end() be true or false?

vzqzhang: > Shouldn't this be `itr_stat == status::iterator_ok`?

Updated.

larryk85: False.  You can't materialize the value from either erased or end.

So the use case for this is:
```c++
auto itr = find some value
if (itr)
   do something with the value
else
   create an entry for the key and value
```

vzqzhang:
Since CI fixes are required for buildkite build, I created a new PR: https://github.com/EOSIO/eosio.cdt/pull/997

PR #989, opened 2020-11-05
abigenerator bug fix

vzqzhang:


larryk85 (changes requested):
Just adding the note here for posterity, we have already discussed this.

Please add toolchain tests that test this issue so that we won't regress in the future.

vzqzhang:
> Just adding the note here for posterity, we have already discussed this.
> 
> Please add toolchain tests that test this issue so that we won't regress in the future.

Added a toolchain test

tests/toolchain/abigen-pass/kv_table_key_type.cpp
```diff
@@ -0,0 +1,57 @@
...
+      std::string fullname;
+      uint32_t age;
+
+      std::pair<int, int> a;
+      std::optional<float> b;
+      std::list<double> c;
+      std::vector<int> d;
+      std::variant<int, bool, float> e;
+
+      eosio::non_unique<std::string, uint32_t> non_unique_name;
```
larryk85, line 21: Sorry to ask.  But we are removing the alias `eosio::non_unique` can you change these occurrences in this file to `std::tuple` the rest of your code should work as is (let me know if it doesn't).

larryk85 (changes requested):

tests/toolchain/abigen-pass/kv_table_key_type.cpp
vzqzhang: Updated the test. It works.

larryk85 (approved):
Please wait until everything is green before merging.  Thanks again.

PR #990, opened 2020-11-05
docs/kv address example updates to fix non-unique indexes range scenario.

iamveritas:
Updates to the kv-address-book example smart contract:

* got rid of the `using namespace eosio;` and `using namespace std;` and use it explicitly everywhere else to help the user see better from what namespace each object comes from; we assume the users reading this example are beginners and need this explicit style of coding
* fix the non-unique indexes to be able to showcase the `get all persons in the addressbook with the same last name` scenario.


examples/kv_addr_book/include/kv_addr_book.hpp
```diff
@@ -58,8 +56,8 @@ struct person_factory {
       std::string personal_id) {
          return person {
             .account_name = account_name,
-            .first_name = {account_name, first_name},
-            .last_name = {account_name, last_name},
+            .first_name = {first_name, last_name},
```
kimjh2005, line 59: why do you add last_name to first_name?

iamveritas: it is how currently the non-unique indexes work.
in order to call a `range()` for `last_name_idx` non-unique index with a fixed value for last name, you have to do it this way, declare the first name as a eosio::non_unique tuple

kimjh2005 (approved):

iamveritas:
this can not be merged yet, because we can't yet get rid of the using namespace statements

PR #996, opened 2020-11-10
KV fixes

larryk85:
## Change Description

This PR fixes some issues with KV and the up port to llvm9. 
1) Issues with building on the most recent versions of clang and gcc.
2) WASM size bloat because of std::function and performance concerns with its use.
3) Cleanup some of the macros.
4) Fix issues with memory management and over allocating.
5) remove non_unique as it isn't truly an index and the abstraction isn't fully thought out or functional.

## API Changes
- [ ] API Changes

The type alias `non_unique` is no longer available.

## Documentation Additions
- [ ] Documentation Additions


linhuang-blockone (approved):

larryk85:
> I cannot find where `cc` is used. Where is `cc` freed?

Yup, that was an oversight (it's a dead variable, thanks).

larryk85:
> should `i=kt.size()` be `i=kt.size() - 1`?

Yeah, I also found it when I actually wrote the test for upper bound, ha.

larryk85:
> Why is `!=` changed to `==`?

Ha ha ha.  Because you were not supposed to see that.

Actually it was a test to see if you were paying attention and you did splendid!

But, really, I hit the undo too much after going down a bit of a rabbit hole with trying to get the writable_wrapper integrated with the iterators (that will have to come later).  You caught the 'boo boo' that the iterator operator== was wrong for a certain condition of iterators.  

Nonetheless, I will now go walk away shamefully.

PR #1002, opened 2020-11-27
Improvements to print overloads and add support for printing std::string_view

arhag:
## Change Description

This PR adds more restrictive conditions for matching the general template `print` overload that is meant to work with classes (i.e. meant to call a `print` member function of the class). It now only enables the overload if the class has a member function called `print` which can take no arguments. To continue supporting `std::string` (which was previously added via `if constexpr` within this general template overload for some reason), a new explicit overload of the `print` function has been added. In addition, another explicit overload of the `print` function has been added to also support `std::string_view`.

@larryk85 Please consider including this patch into the `develop` branch as well.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/core/eosio/print.hpp
+++ b/libraries/eosiolib/core/eosio/print.hpp
@@ -45,24 +45,6 @@ namespace eosio {
          void printhex(const void*, uint32_t);
       }
 
-
-      template <typename U>
-      struct has_print_member_function
-      {
-      private:
-         template <typename T>
-         struct drop {};
-
-         template <typename T>
-         static auto test(int)
-            -> std::enable_if_t<(drop<decltype(std::declval<T>().print())>{}, true),
-                                std::true_type>;
-
-         template <typename>
-         static auto test(...) -> std::false_type;
-      public:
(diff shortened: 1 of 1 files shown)


libraries/eosiolib/core/eosio/print.hpp
```diff
@@ -160,16 +199,14 @@ namespace eosio {
...
-         internal_use_do_not_use::prints_l( t.c_str(), t.size() );
-      else if constexpr (std::is_same<std::decay_t<T>, char*>::value)
-         internal_use_do_not_use::prints(t);
-      else
-         t.print();
+   template <typename T, std::enable_if_t<
+                           internal_use_do_not_use::has_print_member_function<
+                              std::decay_t<T>>::value,
+                           int> = 0>
+   inline void print(T&& t) {
```
swatanabe-b1, line 208: You could just use `auto print(T&& t) -> std::void_t<decltype(t.print())>`

swatanabe-b1 (approved):

PR #1013, opened 2020-12-08
KV map fixes

larryk85:
## Change Description

During some dogfooding by b1bart.  We uncovered a few simple bugs.  These are the quick fixes and associated tests that should've been there to begin with and some to ensure we don't regress.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/eosio.hpp
+++ b/libraries/eosiolib/contracts/eosio/eosio.hpp
@@ -9,7 +9,6 @@
 #include "multi_index.hpp"
 #include "dispatcher.hpp"
 #include "contract.hpp"
-#include "../../core/eosio/reflect.hpp"
 
 #ifndef EOSIO_NATIVE
 static_assert( sizeof(long) == sizeof(int), "unexpected size difference" );
--- a/libraries/eosiolib/contracts/eosio/eosio.hpp
+++ b/libraries/eosiolib/contracts/eosio/eosio.hpp
@@ -9,6 +9,7 @@
 #include "multi_index.hpp"
 #include "dispatcher.hpp"
 #include "contract.hpp"
+#include "../../core/eosio/reflect.hpp"
 
 #ifndef EOSIO_NATIVE
 static_assert( sizeof(long) == sizeof(int), "unexpected size difference" );
--- a/libraries/eosiolib/contracts/eosio/map.hpp
+++ b/libraries/eosiolib/contracts/eosio/map.hpp
@@ -431,7 +431,8 @@ namespace eosio::kv {
          }
 
          inline bool empty() const {
-            auto it = lower_bound("");
+            iterator_t it = {owner};
+            it.lower_bound(key_type{""});
             return it == end();
          }
 
@@ -498,6 +499,19 @@ namespace eosio::kv {
 
          friend iterator_t;
 
+         static detail::packed_view get_tmp_buffer(std::size_t size_needed=0) {
+            constexpr std::size_t max_size = 512;
+            static char static_data[max_size];
+            static std::vector<char> dynamic_data = {0};
+            if (size_needed > max_size) {
+               if (dynamic_data.size() < size_needed)
+                  dynamic_data.resize(size_needed);
+               return {dynamic_data.data(), size_needed};
+            } else {
+               return {&static_data[0], size_needed};
+            }
+         }
+
       protected:
          constexpr static inline name index_name = name{IndexName};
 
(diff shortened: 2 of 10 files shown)


tests/unit/test_contracts/kv_map_tests.cpp
```diff
@@ -138,4 +139,45 @@ class [[eosio::contract]] kv_map_tests : public eosio::contract {
...
+   void constrct() {
+      using map_t = eosio::kv::map<"map"_n, float, eosio::time_point>;
+      map_t m = {{13.3f, eosio::time_point{}}};
+
+      m[13.3f] = eosio::time_point{eosio::microseconds{(int64_t)13}};
+
+      auto iter = m.find(13.3f);
+      auto val = iter->second();
+      auto expected = eosio::time_point{eosio::microseconds{(int64_t)13}};
+      eosio::check(val == expected, "should be equal and not fail to compile");
```
linhuang-blockone, line 180: Can you also check the size of m is 1, and m.find(12.0f) returns null.

linhuang-blockone (approved):

libraries/eosiolib/contracts/eosio/eosio.hpp
```diff
@@ -9,6 +9,7 @@
 #include "multi_index.hpp"
 #include "dispatcher.hpp"
 #include "contract.hpp"
+#include "../../core/eosio/reflect.hpp"
```
b1bart, line 12: Do we have a guideline for what does and does not go into `eosio/eosio.hpp`

This seems like something more appropriately pushed down into the headers (like map.hpp or producer_schedule.hpp). It seems that at least in this header map.hpp precedes the inclusion of reflect.hpp and map.hpp uses the CDT_REFLECT macro.

b1bart (changes requested):

tests/unit/test_contracts/kv_map_tests.cpp
larryk85: Sure.

libraries/eosiolib/contracts/eosio/eosio.hpp
larryk85: We do.  It is something that would be used by people in a lot of cases.  Historically, eosio.hpp is supposed to be a catchall header for a one header-include for the majority of normal operations.  After this release, the old EOSLIB_SERIALIZE macro is just going to be replaced full sail with the CDT_REFLECT.  They are fundamentally different in how they work, but allow for the same things to be done, but CDT_REFLECT allows for a lot more to be done on custom data types.  Users might have to use it with their custom structs for constructing a key, it is determined by a few separate factors (i.e. PFR fails because it is not a POD type and those types of situations.

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -498,6 +499,19 @@ namespace eosio::kv {
 
          friend iterator_t;
 
+         static detail::packed_view get_tmp_buffer(std::size_t size_needed=0) {
```
b1bart, line 502: moving this helped `::materialize` compile to the next error which only occurs when using accessing `const` iterator:

```
/include/eosiolib/contracts/eosio/map.hpp:262:13: error: 'this' argument to member function 'materialize' has type 'const eosio::kv::detail::iterator<false, eosio::kv::map<...> >', but function is not marked const
            materialize();
            ^~~~~~~~~~~
```

tests/unit/test_contracts/kv_map_tests.cpp
```diff
@@ -138,4 +139,48 @@ class [[eosio::contract]] kv_map_tests : public eosio::contract {
...
+
+   [[eosio::action]]
+   void constrct() {
+      using map_t = eosio::kv::map<"map"_n, float, eosio::time_point>;
+      map_t m = {{13.3f, eosio::time_point{}}};
+
+      m[13.3f] = eosio::time_point{eosio::microseconds{(int64_t)13}};
+
+      auto iter = m.find(13.3f);
+      auto val = iter->second();
```
b1bart, line 178: As it follows a different code path we should have something that dereferences a const iter

b1bart (changes requested):

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -298,7 +298,7 @@ namespace eosio::kv {
 
          inline bool operator!=(const iterator& o) const { return !((*this) == o); }
 
-         void materialize() {
+         void materialize() const {
```
b1bart, line 301: Notes for future versions:
- this will re-materialize the element that you may hold a reference to but only when you dereference the iterator you got it from.  This seems like an odd interaction from the outside.
- perhaps `element` should be a `std::unique_ptr` or other such construct to reduce the size of an iterator when the value is large and to enforce less requirements (copiability/movability/default-constructability etc) on the value type

b1bart (approved):

PR #1023, opened 2020-12-30
[docs] release/1.8.x add a few improvements for docs

### LINKED ISSUE #818: Document how to properly use `multi_index::get` and deprecate related methods that take a reference
We need clear documentation on how to use `multi_index::get` so that contract developers avoid the common pitfall of copy assigning the reference returned from `multi_index::get` to a stack-allocated local variable and then passing that into `multi_index::modify`.

The most common mistake is that users use `auto` as the typename for the local variable when they should be using `auto&` or `decltype(auto)`.

Also, because this is a common mistake contract developers make which could lead to undefined behavior, we should deprecate the overloads of `modify` and `erase` methods in `multi_index` that take a reference, and encourage them to use the versions that take `const_iterator` instead which are much safer.

We should also provide warnings in the documentation about using `iterator_to` which can have similar undefined behavior if the developer passes in a reference to a stack-allocated object rather than the reference returned by `get` or by dereferencing a `const_iterator`. Perhaps we can also provide a safer overload of `iterator_to` that takes a `const_iterator`.

Along with deprecating `multi_index::get`, we should also make sure we do not actually use it in any of the contracts in our repository. Instead we could use `multi_index::require_find` from EOSIO/eos#4227.

iamveritas:
this is a sister PR for https://github.com/EOSIO/eosio.cdt/pull/1028

a few improvements:

- resolves #818, add warning for multi_index::get and multi_index::iterator_to
- correct the link to the built in types 
- and improve the kv_map example with a new action that demonstrates input param as user defined type

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,14 +1,14 @@
 # EOSIO.CDT (Contract Development Toolkit)
 
-## Version : 1.7.0
-
 EOSIO.CDT is a toolchain for WebAssembly (WASM) and set of tools to facilitate smart contract development for the EOSIO platform. In addition to being a general purpose WebAssembly toolchain, [EOSIO](https://github.com/eosio/eos) specific optimizations are available to support building EOSIO smart contracts.  This new toolchain is built around [Clang 7](https://github.com/eosio/llvm), which means that EOSIO.CDT has the most currently available optimizations and analyses from LLVM, but as the WASM target is still considered experimental, some optimizations are incomplete or not available.
 
 ## New Introductions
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -1,14 +1,14 @@
 # EOSIO.CDT (Contract Development Toolkit)
```
iamveritas, line 1: changes in this file are cosmetics mostly

lparisc (approved):

PR #1034, opened 2021-01-20
[docs] attempt to separate annotations from code into a dox file

iamveritas:
this is how the dox file supposed to be built based on doxygen docs and other samples out there
however it is not generating documentation for the functions of the iterator class, for some reason which escapes me it ignores all of them... :(
it does generate the map and iterator class documentation though, but that's it no functions.


iamveritas:
we will not pursue this path, to have annotations separate from code.

PR #1035, opened 2021-01-20
[docs] dev - include c api annotations to reference API generation in cdt

### LINKED ISSUE #819: Document that update intrinsics of C DB API cannot change primary key
The current (as of v1.0.5) update intrinsics of the C DB API do not allow changing the primary key of a table row. The serialized data that is stored in the table row of a primary table may include a primary key and that primary key value could be changed by the contract calling the `db_update_i64` intrinsic. But that does not change the actual primary key recognized by the native side.

This behavior should be well documented for contract developers that use the C DB API. Those using the C++ `multi_index` shouldn't have to worry about this (although it should still be documented) because the `modify` method will check that the primary key was not changed by the `updater` lambda and cause an assertion failure otherwise.

iamveritas:
fixes #819
include c api (*.h files) and tell doxygen to ignore the __attribute__ macros

sister PR: https://github.com/EOSIO/eosio.cdt/pull/1036

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -11,7 +11,11 @@
 			"name": "doxygen_to_xml",
 			"options": {
 				"INPUT": "libraries/eosiolib",
-				"EXCLUDE_PATTERNS": "*.cpp *.c *.h"
+				"EXCLUDE_PATTERNS": "*.cpp *.c",
+				"ENABLE_PREPROCESSING": "YES",
+				"MACRO_EXPANSION": "YES",
+				"EXPAND_ONLY_PREDEF": "YES",
+				"PREDEFINED": "__attribute__(x)="
 			}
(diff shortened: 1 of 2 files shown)


lparisc (approved):
Looks good, just two suggestions:
- consider creating a ticket later to address the double descriptions under each function. These result from having both a standalone description and a redundant \@brief command. Standalone descriptions are more common for functions, while @brief is more common for anything else (classes, namespaces, enums, etc.)
- consider replacing the \@details command with \@remarks instead. The latter shows a "remarks" title which is more appropriate. The former doesn't (because it's expected to expand the description). Currently, it looks as if it's expanding the last \@post command since it has no title.

Also, glad to see the C API back!

PR #1036, opened 2021-01-20
[docs] 1.8.x - include c api annotations to reference API generation in cdt 

### LINKED ISSUE #819: Document that update intrinsics of C DB API cannot change primary key

iamveritas:
fixes #819
include c api (*.h files) and tell doxygen to ignore the attribute macros

sister PR
https://github.com/EOSIO/eosio.cdt/pull/1035

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -11,7 +11,11 @@
 			"name": "doxygen_to_xml",
 			"options": {
 				"INPUT": "libraries/eosiolib",
-				"EXCLUDE_PATTERNS": "*.cpp *.c *.h"
+				"EXCLUDE_PATTERNS": "*.cpp *.c",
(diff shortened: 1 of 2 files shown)


lparisc (approved):
Looks good, just check my comments on PR #1035

PR #1039, opened 2021-01-26
[docs] dev - fix annotations for kv map, kv table and multi index

iamveritas:
- fix code annotations for kv map, kv table and multi index.

- got rid of the `map_iterator.dox` file cause I could not find a way to make it work properly and we do not want to move in the direction of separating annotations from code.

- remove redundant annotations from db.h file

sister PR: https://github.com/EOSIO/eosio.cdt/pull/1040


libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -132,6 +136,15 @@ namespace eosio::kv {
          return internal_use_do_not_use::kv_it_lower_bound(itr, "", 0, _k, _v);
       }
 
+      /**
+       * @ingroup keyvaluemap
+       * 
+       * @brief This struct represents the data type stored in map.
+       * @details You will use the sets of functions and operations associated with this type to
```
halsaphi, line 143: sets or set?

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -157,6 +170,17 @@ namespace eosio::kv {
          name     payer;
       };
 
+      /**
+       * @ingroup keyvaluemap
+       * 
+       * @brief This struct represents the iterator for the `eosio::kv::map` data type.
+       * @details You will use the sets of functions and operations associated with this type to
```
halsaphi, line 177: sets or set ... won't add more comments on each instance of this

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -157,6 +170,17 @@ namespace eosio::kv {
...
+       * iterate through values in the map and reference them.
+       * 
+       * Note that this iterator type in `eosio::kv::map` is defined as `eosio::kv::map::iterator_t`.
+       * There is also a reverse iterator available, it is defined as `eosio::kv::map::reverse_iterator_t`.
+       * The only difference between iterator_t and reverse_iterator_t is the direction the ++ and -- operators go.
```
halsaphi, line 182: The only difference between iterator_t and reverse_iterator_t is the direction the ++ and -- operators go. - would it be better to define what ++ / -- does for each iterator rather than just say they work in opposite ways?

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -216,29 +246,64 @@ namespace eosio::kv {
                itr_destroy(handle);
          }
 
+         /**
+          * @ingroup keyvaluemap
+          *
+          * @brief Utility function to see if the iterator is valid or is pointing 
+          * at `end` or one past the last or first element.
```
halsaphi, line 253: @brief Utility function to see if the iterator is valid or is pointing 
          * at `end` or one past the last or first element.
would 
@ brief Utility function return true if iterator is valid else returns false if iterator is not valid - then define what is not valid
be better?

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -216,29 +246,64 @@ namespace eosio::kv {
...
             current_status = static_cast<status>(itr_move_to_end(handle));
             return *this;
          }
 
+         /**
+          * @ingroup keyvaluemap
+          *
+          * @brief This will take a `key_type` key and find the value that is equal to or greater than that key.
+          * @details If no element is greater than or equal to the key the iterator will now hold `end`.
+          * @param key This is the key which you wish to query with.
```
halsaphi, line 293: @param key This is the key which you wish to query with.
or
@param key The search key.

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -216,29 +246,64 @@ namespace eosio::kv {
...
             current_status = static_cast<status>(itr_lower_bound(handle, {k.data(), k.size()}));
             return *this;
          }
 
+         /**
+          * @ingroup keyvaluemap
+          *
+          * @brief This will take a `key_type` key and find the value that exactly matches that key.
+          * @details If no element is equal to the key the iterator will now hold `end`.
+          * @param key This is the key which you wish to query with.
```
halsaphi, line 305: * @param key This is the key which you wish to query with.
 or 

 @param key The search key  ... or maybe the query key

won't comment on this one again

halsaphi: + is it worth mentioning these two return a reference rather than the iterator as opposed to the lower_bound queries below

libraries/eosiolib/contracts/eosio/map.hpp
```diff
@@ -457,12 +562,26 @@ namespace eosio::kv {
...
             return it;
          }
 
+         /**
+          * @ingroup keyvaluemap
+          *
+          * @brief This will take a `key_t` key and find the value that is strictly greater than that key.
+          * @details If no element is strictly greater than the key the iterator will now hold `end`.
+          * @param key This is the key which you wish to query with.
+          */
```
halsaphi, line 584: param description as above + return type

libraries/eosiolib/contracts/eosio/table.hpp
```diff
@@ -393,17 +400,16 @@ namespace internal {
...
- *
+ * @ingroup keyvaluetable
+ * 
  * @brief Defines an EOSIO Key Value Table
  * @details EOSIO Key Value API provides a C++ interface to the EOSIO Key Value database.
  * Key Value Tables require 1 primary index, of any type that can be serialized to a binary representation.
  * Key Value Tables support 0 or more secondary index, of any type that can be serialized to a binary representation.
  * Indexes must be a member variable or a member function.
  *
  * @tparam T         - the type of the data stored as the value of the table
```
halsaphi, line 411: - the type of the data stored as the value of the table
or 
- the type of the data stored as a value in the table
?

libraries/eosiolib/contracts/eosio/map.hpp
iamveritas: I have no idea... I guess both work?... no?

iamveritas: I will change it to set anyway (these are the dev's comments and I did not want to interfere too much with them content unless it was obvious something not clear/good)

libraries/eosiolib/contracts/eosio/map.hpp
iamveritas: this note is at the iterator class level; I do not see a problem to have it here.
indeed the -- and ++ should be annotated as well separatelly.

libraries/eosiolib/contracts/eosio/map.hpp
iamveritas: true

libraries/eosiolib/contracts/eosio/map.hpp
iamveritas: I have an open thread with the dev on why it is not returning a reference... it is about the interator_t vs iterator (two different things)

iamveritas: as for "The query key" I like more what the dev did, it is more explicit. What you are proposing it is shorter in deed but it is not so explicit.

lparisc (approved):

linhuang-blockone (approved):

PR #1046, opened 2021-02-03
Create Public Docker Images

kj4ezj:
## Change Description

From [AUTO-457](https://blockone.atlassian.net/browse/AUTO-457), we are working on providing the EOSIO community minimal docker images via Docker Hub which they can pull to more easily use our tooling in a clean environment. We have already introduced a docker image for EOSIO, so this pull request creates one for EOSIO.CDT.

## API Changes
- [ ] API Changes

None.

## Documentation Additions
- [x] Documentation Additions

Public docker images are now available at docker.io/eosio/eosio.cdt.


scottarnette (approved):

.cicd/pipeline.yml
```diff
@@ -421,6 +421,14 @@ steps:
     timeout: 20
     skip: ${SKIP_MACOS_10_15}${SKIP_PACKAGE_BUILDER}
 
+  - wait
+  
+  - label: ":docker::ubuntu: Docker - Build 18.04 Docker Image"
+    command: "./.cicd/create-docker-from-binary.sh"
+    agents:
+      queue: "automation-eks-eos-builder-fleet"
```
kj4ezj, line 429: The `automation-eks-eos-builder-fleet` is a beefier fleet than what we need to run aptitude. Let's switch to the `automation-eks-eos-tester-fleet`, which is a lighter instance type.

docker/dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:18.04
+
+COPY *.deb /
+
+RUN apt update && \
+    apt install -y /*.deb && \
+    rm -rf /*.deb /var/lib/apt/lists/*
```
kj4ezj, line 7: A couple of changes in this code block.
1. First, `apt` is not the same as `apt-get`. The `apt-get` command is intended for scripts and the command-line interface will not be changed, whereas `apt` is meant for humans and should not be used by scripts.
2. Second, we are missing `apt-get clean` to remove cache files before committing this layer.
3. Third, the POSIX standard requires that text files end with a newline `\n` character.

Since these docker images are meant for public consumption, instead of internal consumption, we have to be picky about these types of things and maintain compatibility with common open standards like POSIX. So, this code block...
```dockerfile
RUN apt update && \
    apt install -y /*.deb && \
    rm -rf /*.deb /var/lib/apt/lists/*
```
...would become:
```dockerfile
RUN apt-get update && \
    apt-get install -y /*.deb && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

```

kj4ezj (commented):
Since we're making these other changes anyways, I'd also recommend removing the blank lines. They serve no purpose. But, that is up to you.

.cicd/pipeline.yml
cocho-gore: Done.

docker/dockerfile
cocho-gore: Done. I'll keep in mind thanks

PR #1066, opened 2021-03-11
Add  new intrinsic register_transaction_hook

victorj8:
## Change Description

A new privileged intrinsic with name register_transaction_hook is defined to register a contract callback at the transaction pre-execution phase for all transactions on the chain. 

This will give the system contract the ability to inspect / make modifications to the transaction prior to its execution.  
It is possible that in the future other hook types will be defined.

The interface for the intrinsic is:

`bool register_transaction_hook(transaction_hook hook_type, name callback_contract, name callback_action);`  

where
hook_type, is the type of hook we want to register (currently only one type - preexecution)
callback_contract, is the name of the contract containing the action for the callback
callback_action, is the name of the action inside the contract containing the code for the callback

## API Changes
- [x] API Changes

The register_transaction_hook intrinsic is added.

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/eosiolib/capi/eosio/privileged.h
+++ b/libraries/eosiolib/capi/eosio/privileged.h
@@ -84,6 +84,16 @@ __attribute__((eosio_wasm_import))
 void set_privileged( capi_name account, bool is_priv );
 
 /**
+ * Register a transaction hook (actions that execute at certain points on the lifetime of the execution of a transaction)
+ *
+ * @param hook_type - type of the hook to register
+ * @param callback_contract - name of the contract associated with the hook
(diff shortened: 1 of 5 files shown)


victorj8:
@scottarnette @kj4ezj The integration test is failing, if my understanding is correct, the integration test pulls the docker image from develop branch, however if a new intrinsic is defined then it will cause this test to fail because the intrinsic is not defined on the develop branch, let me know if you think the integration test code should be modified.

larryk85 (approved):
I don't see any issues with this.  But, this PR will need to wait on its counterpart in eos before we can merge this and verify the tests work.

PR #1068, opened 2021-03-16
Enforce read-only nature of an action with read-only attribute

bogniq:
## Change Description

This PR adds checking of `eosio:read_only` attribute for actions. Currently, it only supports checking for direct calls and some ways of indirect calls of specific write host functions. 

If compiler option `--warn-action-read-only` is used, even if some read-only action calls write host functions, the contract compilation still continues but will output warning messages. Otherwise, the compilation will stop and output error messages.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tools/include/eosio/gen.hpp
```diff
@@ -728,19 +738,71 @@ struct generation_utils {
...
+      };
+
+      if (eosio_wasm_import_write_funcs.count(func_decl->getQualifiedNameAsString()) == 0) {
+         return false;
+      }
+
+      if (func_decl->isInExternCContext()) {
+         auto attrs = func_decl->getAttrs();
+         for (auto const &a : attrs) {
+            if (a->getSpelling() == "eosio_wasm_import") {
```
jgiszczak, line 799: This needs to be an explicit string comparison function call.  There doesn't seem to be an appropriate equality operator overload.

jgiszczak (changes requested):

tools/include/eosio/gen.hpp
bogniq: Thanks. `getSpelling` returns a `const char*` but I wrongly thought it returns a std::string. It's fixed in https://github.com/EOSIO/eosio.cdt/blob/4ff779217c54dccc8fb9a59463c0e1c6cb78e82b/tools/include/eosio/gen.hpp#L799

.gitmodules
```diff
@@ -9,6 +9,7 @@
 [submodule "eosio_llvm"]
 	path = eosio_llvm
 	url = https://github.com/eosio/llvm
+	branch = qy-read-only-action
```
larryk85, line 12: This branch should be the 'eosio' branch of llvm.  So, please make a PR to LLVM to the branch 'eosio' and then update this.

tools/include/eosio/codegen.hpp
```diff
@@ -265,16 +281,173 @@ namespace eosio { namespace cdt {
...
+                        process_assignment(bo);
+                     }
+                  }
+               }
+            }
+         }
+
+         virtual bool VisitFunctionDecl(FunctionDecl* func_decl) {
+            SourceManager &sm = get_rewriter().getSourceMgr();
+            if (sm.isInSystemHeader(func_decl->getLocation()) || sm.isInExternCSystemHeader(func_decl->getLocation())) {
```
larryk85, line 385: This is a fine short circuit.  But, we do have a few places in our C++ and C headers where we have some `extern C <writable_host_functions>` to supply the underlying bases for libc.  This will more than likely miss those calls.

tools/include/eosio/gen.hpp
```diff
@@ -683,5 +684,124 @@ struct generation_utils {
 
       return in_kv_namespace && is_internal;
    }
+
+   inline bool is_write_host_func( const std::string& t ) {
+      static const std::set<std::string> write_host_funcs =
+      {
```
larryk85, line 690: I would recommend only using the C name for looking for these host functions.
I.e. instead of `eosio::internal_use_do_not_use::set_resource_limits`, just look for the stripped name of `set_resource_limits` and if it is a extern-C function or a regular C function.

The reason for this, is that as you can see, you have to duplicate some of these, also the C++ namespaced names are purely decorative and just supplying context for the smart contract developer.  You should still only have one `set_resource_limits` per smart contract.

larryk85 (changes requested):

.gitmodules
bogniq: https://github.com/EOSIO/llvm/pull/29 was just created, which updates clang to the head of its 'eosio' branch (with `read-only` attribute added). The `branch` setting here will be removed after the LLVM PR is merged.

tools/include/eosio/codegen.hpp
bogniq: This system header checking is deleted in the new commit: https://github.com/EOSIO/eosio.cdt/blob/ef00c9d035fcea306131884724691398189be4a3/tools/include/eosio/codegen.hpp#L383-L384

tools/include/eosio/gen.hpp
bogniq: Fixed. (Method `is_eosio_wasm_import_write_func` is also removed, because when using C name to find write host functions, the updated `is_write_host_func` method can also handle the "eosio_wasm_import" functions. Please let me know if I missed anything. Thanks!)
https://github.com/EOSIO/eosio.cdt/blob/ef00c9d035fcea306131884724691398189be4a3/tools/include/eosio/gen.hpp#L688-L726

larryk85 (approved):

jgiszczak (approved):

PR #1082, opened 2021-03-30
Adding eosio structure attribute

venu-block1:
## Change Description

Adding eosio::structure custom attribute. Can be used for retrieving metadata. Example usage: struct [[eosio::structure]] fee_payer { /* fields */ };

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

This PR is part of Transaction Sponsorship feature

### DIFF
--- a/tools/include/eosio/abigen.hpp
+++ b/tools/include/eosio/abigen.hpp
@@ -257,6 +257,31 @@ namespace eosio { namespace cdt {
          _abi.tables.insert(t);
       }
 
+      void add_structure( const clang::CXXRecordDecl* decl ) {
+         structures.insert(decl);
+         abi_structure t;
+         t.type = decl->getNameAsString();
+         auto structure_name = decl->getEosioStructureAttr()->getName();
+         if (!structure_name.empty()) {
+            validate_name( structure_name.str(), [&](auto s) { CDT_ERROR("abigen_error", decl->getLocation(), s); } );
+            t.name = structure_name.str();
+         }
+         else {
+            t.name = t.type;
+         }
+         cstructures.insert(t);
+      }
+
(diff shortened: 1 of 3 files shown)


tools/include/eosio/abigen.hpp
```diff
@@ -257,6 +257,31 @@ namespace eosio { namespace cdt {
          _abi.tables.insert(t);
       }
 
+      void add_structure( const clang::CXXRecordDecl* decl ) {
```
larryk85, line 260: I'm a bit confused about this part.

I thought the idea of the new attribute was to flag data types to be included in the 'structs' section for abi conversions to be performed on it.

I don't see why we need another section that is duplicating the information.

larryk85 (changes requested):
I don't see the changes for clang included in this PR.

tools/include/eosio/abigen.hpp
venu-block1: I noticed that eosio::table attribute is made available in both 'structs' section, and also in a separate section, and I assumed that different tools get that information from different sections. I was mimicking that implementation. I can remove the separate section, if my assumption is incorrect, leaving eosio::structure attribute only in 'structs' section. Please advice.

venu-block1:
I was waiting for https://github.com/EOSIO/clang/pull/40 to be approved, so I can  update the submodule changes to this PR. @larryk85, can you please approve it. Thanks.

victorj8:
The file structure_test.json is missing.

tools/include/eosio/abigen.hpp
venu-block1: Removed ABI generation of structures section avoiding duplication.

PR #1098, opened 2021-05-12
Fix the `-v` option of eosio-cpp and eosio-cc to use clang-9

bogniq:
Change `clang-7` to `clang-9` that is called by eosio-cpp and eosio-cc for option `-v`.
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/cc/eosio-cc.cpp.in
+++ b/tools/cc/eosio-cc.cpp.in
@@ -22,7 +22,7 @@ int main(int argc, const char **argv) {
    // fix to show version info without having to have any other arguments
    for (int i=0; i < argc; i++) {
      if (argv[i] == std::string("-v")) {
-       eosio::cdt::environment::exec_subprogram("clang-7", {"-v"});
+       eosio::cdt::environment::exec_subprogram("clang-9", {"-v"});
        return 0;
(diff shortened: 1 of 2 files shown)


dimas1185 (approved):

nksanthosh (commented):
Please also check if clang-x (here 9) binary is shipped with brew or apt (debian package) and RPM package etc. otherwise this wont work

nksanthosh (commented):
You can find details for last stable release for which we uploaded binaries to brew (macOS) and github (for debian and RPM) at https://github.com/EOSIO/eosio.cdt/tree/v1.8.0-rc2#binary-releases - if these work by having clang-7 in them then when we release stable of 1.8.0 it will include the clang-9 binary in it as well. Otherwise we need to check. if we can add it before release

bogniq:
@nksanthosh I checked Ubuntu 18.04, CentOS 8, and MacOS 10.15, and all have clang-7 after installing the 1.7.0 stable release. (Both Ubuntu and CentOS: /usr/opt/eosio.cdt/1.7.0/bin/)

PR #1106, opened 2021-05-22
Fix 128-bit integer operations in RT library

bogniq:
## Change Description

The following `__int128` related code are ported from LLVM `compiler-rt` to implement the related operations in `libraries/rt` : ashlti3, ashrti3, lshrti3, multi3, divti3, udivti3, modti3, and umodti3.

Major changes to the ported `compiler-rt` code include: header file change (e.g. `int_t.h`), name changes to avoid conflicts (e.g. from `__multi3` to `___multi3`), and using wrappers of `libraries/rt` functions in the test code.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/tester/tester.cpp
```diff
@@ -146,3 +146,19 @@ TEST_CASE_METHOD(eosio::test_chain, "Creating signatures", "[sign]") {
    auto r1_sig = eosio::sign(r1_pvt, hash);
    as("test"_n).act<tester_tests::assertsig_action>(hash, r1_sig, r1_pub);
 }
+
+TEST_CASE("Runtime library 128-bit integer operations") {
+   volatile int128_t a = -100, b = 2;
```
huangminghuang, line 151: Why do you need to declare the variable `volatile`? It only makes sense when the variable is used to communicate with a signal handler which is definitely not the case here.

bogniq: Here `volatile` keyword is to ensure compiler won’t optimize out these variables (in CMakeLists.txt, `-Os` option is used to compile `tester.cpp`); otherwise, compiler optimization will cause the int128 functions defined in `rt` library not be called.

huangminghuang (approved):

PR #1116, opened 2021-06-03
[docs] add new ABI eosio::structure attribute

iamveritas:
add new ABI eosio::structure attribute


iamveritas:
not needed anymore, the feature has changed a bit.

PR #1109, opened 2021-05-27
[docs] How Tos update based on new template/format

iamveritas:
update the remaining how tos to conform to our new format/template

other touchups for consistency:
1. multi_index -> multi-index
2. multi index -> multi-index
3. ...

sister PR: https://github.com/EOSIO/eosio.cdt/pull/1117

PR #1117, opened 2021-06-03
[docs] 1.8.x migrate multi-index how tos to the new template

iamveritas:
migrate multi-index how tos to the new template

sister PR: https://github.com/EOSIO/eosio.cdt/pull/1109


iamveritas:
dropped
we will have it done for 2.2

PR #1122, opened 2021-06-21
updated abieos to the latest eosio-2.1.x branch

Max-B1:
I removed usage of `eosio::opaque::unpack()` as it was deprecated in `abieos`.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/eosiolib/contracts/eosio/key_value.hpp
+++ b/libraries/eosiolib/contracts/eosio/key_value.hpp
@@ -170,7 +170,7 @@ inline partial_key make_key(T&& t) {
    return partial_key(convert_to_key(std::forward<T>(t)));
 }
 inline partial_key make_key(partial_key&& t) {
-   return t;
+   return std::move(t);
 }
 inline partial_key make_key(partial_key& t) {
    return t;
--- a/libraries/eosiolib/tester/tester.cpp
+++ b/libraries/eosiolib/tester/tester.cpp
@@ -365,10 +365,10 @@ void build_history_result(eosio::test_chain::get_history_result& history_result,
       history_result.block = std::move(*blocks_result.block);
    }
    if (!blocks_result.traces.empty()) {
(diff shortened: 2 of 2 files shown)


Max-B1:
@heifner @huangminghuang please review

heifner (approved):

libraries/eosiolib/contracts/eosio/key_value.hpp
```diff
@@ -170,7 +170,7 @@ inline partial_key make_key(T&& t) {
    return partial_key(convert_to_key(std::forward<T>(t)));
 }
 inline partial_key make_key(partial_key&& t) {
-   return t;
+   return std::move(t);
```
huangminghuang, line 173: As far as understand, you shouldn't use `std::move(t)` here. You can check https://stackoverflow.com/questions/14856344/when-should-stdmove-be-used-on-a-function-return-value for explanations about when you should (or should not) use `return std::move(...);`

Max-B1: I'm not quite sure about the purpose of these function. The old version of it creates a copy of `t`. So I thought it will make more sense to make it move the content of `t`.
GCC is complaining about it:
```
/Users/maxim.kirov/projects/eosio/eosio.cdt/b1x/libraries/eosiolib/contracts/eosio/key_value.hpp:173:11: warning: local variable 't' will be copied despite being returned by name [-Wreturn-std-move]
   return t;
          ^
/Users/maxim.kirov/projects/eosio/eosio.cdt/b1x/libraries/eosiolib/contracts/eosio/key_value.hpp:173:11: note: call 'std::move' explicitly to avoid copying
   return t;
          ^
          std::move(t)
```

heifner: I believe `move` is appropriate here.

PR #1142, opened 2021-06-29
[develop] Use Anka base image for CICD

scottarnette:
## Change Description

Adding support for Anka base images in CICD, similar to how our CICD in EOSIO/eos works:
- Migrated from `docker` to `platforms`
- Added pre-command support to MacOS steps.
  - This will check for an existing VM to use for build/test steps. If one with the expected dependencies does not already exist, one will be created and uploaded to the Anka registry.
- Added generate-pipeline.sh script.
  - This allows us to check the hash of the MacOS platform scripts and dynamically substitute those values in the pipeline.
  - We'll also support the old static pipeline file for now by checking if this new file exists in your branch before trying to use it.
- Added platform level `SKIP_MAC` flag to skip all MacOS build/test/packing steps.
- Added pre-execute network connectivity checks to all Anka steps.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


kj4ezj (changes requested):
Can we please call the script `pipeline-upload.sh` instead of `generate-pipeline.sh`?
Every repo except [eos](https://github.com/EOSIO/eos) follows this convention.

williamblevins (approved):

kj4ezj (approved):
🍎

PR #1176, opened 2021-07-21
get wasm size artifact

shahabi8:
## Change Description
This pull request creates a Json file artifact to track the wasm and abi files sizes in EOSIO.CDT cicd pipeline. The pipeline is modified such that it builds eosio.contracts to monitor wasm and abi file sizes for system contracts. For Ubuntu 18.04, unit-test step for eosio.contract is included to make sure generated  wasm and abi files are working properly. This is the PR link to EOSIO.CDT release/1.8.x https://github.com/EOSIO/eosio.cdt/pull/1191

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


nksanthosh (changes requested):
A few questions/suggestions:
1. The actual built contracts are not available in the build.tar.gz uploaded to the build step in buildkite. We would want them to be there under test/unit/test_contracts/eosio.contracts (please retain the generated  wasm/abi and sources
2. the wasm_size.log should be generated in the test/unit/test_contracts/ folder and not the root folder unless some consideration requires it and it should reflect the correct path for the various wasm and ABI files. Both wasm and abi file sizes can be tracked
3. Has the automation protocol to pull the wasm_size artifact and graph it (and integrate that graph with the test failure review system been considered in this design?) This can be a separate PR but wanted the design to be supportive of that goal. 
4. How do we do sanity checking that the generated wasms are functional and not broken? Can some smoke tests be added?
5. Please also report the additional time required for the entire CDT build to complete and pass all tests after these changes.

nksanthosh (changes requested):
These changes should be verified via both 
1. the build script 
2. cmake .. and make -j

shahabi8:
> A few questions/suggestions:
> 
> 1. The actual built contracts are not available in the build.tar.gz uploaded to the build step in buildkite. We would want them to be there under test/unit/test_contracts/eosio.contracts (please retain the generated  wasm/abi and sources
> 2. the wasm_size.log should be generated in the test/unit/test_contracts/ folder and not the root folder unless some consideration requires it and it should reflect the correct path for the various wasm and ABI files. Both wasm and abi file sizes can be tracked
> 3. Has the automation protocol to pull the wasm_size artifact and graph it (and integrate that graph with the test failure review system been considered in this design?) This can be a separate PR but wanted the design to be supportive of that goal.
> 4. How do we do sanity checking that the generated wasms are functional and not broken? Can some smoke tests be added?
> 5. Please also report the additional time required for the entire CDT build to complete and pass all tests after these changes.

1. The whole build directory of eosio.contracts is added under `test/unit/test_contracts/eosio.contracts `path.
2. Done.
3. My initial thought is to create a Python script to download wasm artifacts over time and create a graph for each wasm and abi file. I'll create separate PR for it.
4. EOSIO.CONTRACT unit-test is added to Ubuntu 18.04 build for sanity check.
5. Additional execution time for EOSIO.CONTRACT build is reported in the log file. Note that for Ubuntu 18.04 this additional time includes the time to build and execute unit-tests.

shahabi8:
> These changes should be verified via both
> 
> 1. the build script
> 2. cmake .. and make -j

I used cmake commands to build EOSIO.CONTRACT, do you mean I should build EOSIO.CONTRACT both ways?

bogniq:
> > These changes should be verified via both
> > 
> > 1. the build script
> > 2. cmake .. and make -j
> 
> I used cmake commands to build EOSIO.CONTRACT, do you mean I should build EOSIO.CONTRACT both ways?

@nksanthosh For the purpose of getting WASM & ABI sizes, I think the current way (cmake & make) is sufficient, because they are what `build.sh` essentially does (except script option handling).

.cicd/build.sh
```diff
@@ -45,5 +65,71 @@ else # Linux
...
+fi
+
+ELAPSED_TIME=$(($(date +%s)-$START_TIME))
+
+if [[ $BUILDKITE == true ]]; then
+    cd $BUILD_DIR_PATH/tests/unit/test_contracts
+    touch wasm_abi_size.log
+    PATH_WASM=$(pwd)
+    echo '--- :arrow_up: Generating wasm_abi_size.log file'
+    echo '####### EOSIO test contracts wasm files sizes #######'>> $PATH_WASM/wasm_abi_size.log
```
bogniq: What is the design consideration about this log file's structure, so as to facilitate next work of parsing these files to generate time-range report?

bogniq (commented):
LGTM. Some comments about the log file structure.

.cicd/build.sh
shahabi8: The log file has the names and paths of abi and wasm files and their sizes, I can also add date for simplicity of parsing and generating time-range report.

praphael: You might want to just make it JSON.  Would make parsing easier.

bogniq: @kj4ezj Could you provide insight about the log file format and check this PR from CICD perspective? Thanks!

kj4ezj: Yes, JSON artifacts should be used so that a lambda can scoop the artifact and funnel it into Datadog or wherever else you want these metrics to go.

You can use `jq` to generate a JSON artifact. Documentation is [here](https://stedolan.github.io/jq/manual).

.cicd/build.sh
```diff
@@ -3,18 +3,29 @@ set -eo pipefail
 . ./.cicd/helpers/general.sh
 
 mkdir -p $BUILD_DIR
+CDT_DIR_PATH=$(pwd)
+BUILD_DIR_PATH=$CDT_DIR_PATH/$BUILD_DIR
```
kj4ezj: The `CDT_DIR_PATH` and `BUILD_DIR_PATH` variables are duplicates of `ROOT_DIR` and `BUILD_DIR`, respectively, brought in from `.cicd/hepers/general.sh` [here](https://github.com/EOSIO/eosio.cdt/blob/1887998b8092110b5d64c97a56cc48c089d48a5f/.cicd/helpers/general.sh#L1). Let's use those.

.cicd/build.sh
```diff
@@ -3,18 +3,29 @@ set -eo pipefail
...
+BIN_DIR_PATH=$BUILD_DIR_PATH/bin
+git clone -b develop https://github.com/EOSIO/eosio.contracts.git
+CONTRACTS_DIR_PATH=$(pwd)/eosio.contracts
```
kj4ezj: ~~Since you guys are integrating [eosio.contracts](https://github.com/EOSIO/eosio.contracts.git) into the `build.tar.gz`, these `BIN_DIR_PATH` and `CONTRACTS_DIR_PATH` variables will be useful in other parts of the CI code as well, such as your new metrics step and/or the test steps. Let's move these declarations into the `.cicd/hepers/general.sh` lib brought in by all the CI steps so it is available anywhere.~~ (**Edit:** We decided on a video call that [eosio.contracts]() will _not_ be baked into the `build.tar.gz`, so this comment isn't relevant and the `git` clone can occur in the new WASM size metrics/test step.) We can also simplify the names to match [existing patterns](https://github.com/search?q=org%3AEOSIO+CONTRACTS_DIR&type=code).
```bash
export BIN_DIR="$BUILD_DIR/bin"
export CONTRACTS_DIR="$ROOT_DIR/eosio.contracts"
```

.cicd/build.sh
```diff
@@ -3,18 +3,29 @@ set -eo pipefail
 . ./.cicd/helpers/general.sh
 
 mkdir -p $BUILD_DIR
+CDT_DIR_PATH=$(pwd)
+BUILD_DIR_PATH=$CDT_DIR_PATH/$BUILD_DIR
+BIN_DIR_PATH=$BUILD_DIR_PATH/bin
+git clone -b develop https://github.com/EOSIO/eosio.contracts.git
```
kj4ezj: For the `git clone`, we should make the branch configurable.
```bash
git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git
```
If you really want to, you can add a default value...
```bash
git clone -b "${CONTRACTS_VERSION:-develop}" https://github.com/EOSIO/eosio.contracts.git
```
...but our existing pattern is to define CI dependencies and expectations in a `pipeline.jsonc` file [here](https://github.com/EOSIO/eosio.cdt/blob/1887998b8092110b5d64c97a56cc48c089d48a5f/pipeline.jsonc#L2), in the root of the repo. The new dependency might be declared like this:
```jsonc
{
    "eosio-dot-cdt":
    {
        "pipeline-branch": "master",
        "dependencies": // dependencies to pull for cdt integration tests, by branch, tag, or commit hash
        {
            "eosio": "develop",
            "eosio.contracts": "develop"
        }
    }
}
```
This is a pattern we follow in all of the EOSIO repos. For example, the [eos pipeline file](https://github.com/EOSIO/eos/blob/abcaea5bea22729398eda5c71eabf96619cec402/pipeline.jsonc) has a lot of definitions.

Unfortunately, the [eosio-dot-cdt](https://buildkite.com/EOSIO/eosio-dot-cdt) pipeline was neglected for some time and is not following contemporary patterns, so the values from the `pipeline.jsonc` file are not being brought in automatically for you and some work will need to be done here.

I think the easiest way to get this going will be to just recycle [the code that parses this file](https://github.com/EOSIO/eos-buildkite-pipelines/blob/2dc4fb9e6c303e51e85880f79f58c6b2e2c85cb9/scripts/pipeline-branch.sh#L7) and put the following at the end of `.cicd/hepers/general.sh`:
```bash
[[ -z $PIPELINE_CONFIG ]] && export PIPELINE_CONFIG='pipeline.json'
[[ -z $RAW_PIPELINE_CONFIG ]] && export RAW_PIPELINE_CONFIG='pipeline.jsonc'
if [[ -f "$RAW_PIPELINE_CONFIG" ]]; then
    # parse pipeline config
    cat "$RAW_PIPELINE_CONFIG" | grep -Po '^[^"/]*("((?<=\\).|[^"])*"[^"/]*)*' | jq -c '.[env.BUILDKITE_PIPELINE_SLUG]' > "$PIPELINE_CONFIG"
elif [[ "$DEBUG" == 'true' ]]; then
    echo "WARNING: No pipeline configuration file \"$RAW_PIPELINE_CONFIG\" found in \"$(pwd)\""'!'
fi
```
This will leave a `pipeline.json` behind in the root of the repo, without comments so that can be parsed using `jq` in your scripts:
```bash
[[ ! -z "$CONTRACTS_VERSION" ]] || export CONTRACTS_VERSION="$(cat "$PIPELINE_CONFIG" | jq -r '.dependencies["eosio.contracts"]')"
git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git
```

Finally, it looks like `jq` is not currently in our [CDT base-images](https://github.com/EOSIO/eosio.cdt/tree/1887998b8092110b5d64c97a56cc48c089d48a5f/.cicd/platforms), so you will need to add it. This is fine, since you'll need it to generate the `*.json` metrics artifact anyway.

.cicd/build.sh
```diff
@@ -3,18 +3,29 @@ set -eo pipefail
...
     # You can't use chained commands in execute
-    cd $BUILD_DIR
+    cd $BUILD_DIR_PATH
     cmake .. -DCMAKE_BUILD_TYPE=Release
     make -j$JOBS
+    export PATH=$BIN_DIR_PATH:$PATH
+    cd $BUILD_DIR_PATH/tests/unit/test_contracts
+    mkdir -p eosio.contracts
+    cd eosio.contracts
+    START_TIME=$(date +%s)
```
kj4ezj: > Please also report the additional time required for the entire CDT build to complete and pass all tests after these changes.  
  
I believe @nksanthosh was asking you to report on this once for the purpose of understanding how this pull request impacts our build times, as opposed to reporting this for every build until the end of time. Either way, we already collect metrics on the runtime of Buildkite steps, so this can be removed.

.cicd/build.sh
```diff
@@ -3,18 +3,29 @@ set -eo pipefail
...
+    cmake $CONTRACTS_DIR_PATH 
+    make -j$JOBS
```
kj4ezj: What is the fundamental purpose of this second build? What are we building, and what is it for? Will the outputs of this build be used by the CDT tests, or is this solely for WASM size metrics collection?

This almost certainly does not belong in the CDT build step. If this contract build fails, the CDT build will fail as well, the pull request will say CDT failed to build, and it will be confusing to people that their code builds but the Buildkite step is failing. If the only purpose of this second round of `cmake` and `make` is to build the WASMs so you can collect metrics on the sizes, then this (along with the contracts `git` clone) should be moved into your new WASM metrics Buildkite step. If the tests will be using this code, too, then we should create a second "Contracts Build" job step in Buildkite for each operating system where we pull the CDT `build.tar.gz` from this step, pack in contracts, do this second build, and attach the new `build.tar.gz` as an artifact for the tests to pull down. I'd be happy to collaborate on this if you need help.

.cicd/build.sh
```diff
@@ -45,5 +65,71 @@ else # Linux
     fi
 
     eval docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS\"
+    START_TIME=$(date +%s)
+    eval docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS_EOSIO_CONTRACTS\"
```
kj4ezj: Same on Linux, this should almost certainly be a separate Buildkite job step.

kj4ezj (changes requested):
I don't think the build step is an appropriate place to be adding this code because the metrics collection code has nothing to do with the ability to build CDT.

For example, say the metrics collection code fails with some simple error, like a variable interpolation, iteration, or artifact upload. In that case, Buildkite will report to the pull request that CDT failed to build. If a user clicks on their build and looks at it in Buildkite, it will still look like the CDT build itself failed. This will cause a ton of confusion because, of course, CDT _does_ build, but the build step is failing. We really want the build step to focus exclusively on building.

Instead, let's make a separate Buildkite job step in the pipeline that runs with the tests on the tester fleet, pulls down the `build.tar.gz`, collects your metrics, attaches an artifact like `metrics.json`, `wasm.json`, `wasm-metrics.json`, or similar; then exits. That step could (optionally) have `soft_fail: true` set as well, so it cannot block merges or releases. (**Edit:** We decided against `soft_fail` on a video call.) Then a Lambda can come, scoop the `wasm-metrics.json`, funnel it into Datadog and wherever else, and alert to PagerDuty and/or Slack if the metrics are missing.

I am happy to provide assistance, as-needed.

.cicd/pipeline.yml
```diff
@@ -147,6 +147,150 @@ steps:
...
+  - label: ":aws: Amazon_Linux 2 - EOSIO.CONTRACTS Build"
+    command:
+      - "buildkite-agent artifact download build.tar.gz . --step ':aws: Amazon_Linux 2 - Build' --agent-access-token $$BUILDKITE_AGENT_ACCESS_TOKEN && tar -xzf build.tar.gz"
+      - "./.cicd/eosio-contracts-build.sh"
+    env:
+      IMAGE_TAG: "amazonlinux-2"
+    agents:
+      queue: "automation-eks-eos-tester-fleet"
+    timeout: ${TIMEOUT:-10}
+    skip: ${SKIP_AMAZON_LINUX_2}${SKIP_LINUX}${SKIP_UNIT_TESTS}
```
bogniq: Even `*** - EOSIO.CONTRACTS Build` tasks are in the same step as unit tests, I think it's better to replace `SKIP_UNIT_TESTS` with a new variable like `SKIP_WASM_METRICS` for these tasks, so that it's possible to skip WASM metrics tests and unit tests separately.

nksanthosh:
Please confirm all these variants work

nodeos version = v2.1.0 (exact version test) , release/2.1.x (head of this branch)
Same for contracts
PR to  develop and 1.8.x of CDT

.cicd/build.sh
shahabi8: @kj4ezj I found that in part of the code where I query the wasm sizes ROOT_DIR and BUILD_DIR don't work. So, I defined CDT_DIR_HOST, which is address of CDT source file on the host.

.cicd/build.sh
shahabi8: @kj4ez, from this address CDT_DIR_HOST I could navigate to all other directories needed.

.cicd/build.sh
shahabi8: Done.

.cicd/build.sh
shahabi8: Done.

.cicd/build.sh
shahabi8: @kj4ezj, I modified the pipeline based on this comment.

.cicd/build.sh
shahabi8: Done.

.cicd/pipeline.yml
shahabi8: @bogniq, yes I added SKIP_WASM_METRICS as key to skip all the wasm size related steps in the pipeline.

shahabi8:
> Please confirm all these variants work
> 
> nodeos version = v2.1.0 (exact version test) , release/2.1.x (head of this branch)
> Same for contracts
> PR to develop and 1.8.x of CDT

@nksanthosh, I verified the following variants:
EOSIO/release/2.1.x, EOSIO.CDT/release/1.8.x, EOSIO.CONTRACTS/develop (https://buildkite.com/EOSIO/eosio-dot-cdt/builds/4258#277ecf9b-3ef3-4c40-9e8b-fbc7d8ba2166)
EOSIO/v2.1.0, EOSIO.CDT/release/1.8.x, EOSIO.CONTRACTS/develop (https://buildkite.com/EOSIO/eosio-dot-cdt/builds/4260#9f7bd1de-808f-40d6-9d36-8a58687c8951)

Here is the PR link to CDT 1.8.x: https://github.com/EOSIO/eosio.cdt/pull/1191

.cicd/pipeline.yml
```diff
@@ -594,15 +738,15 @@ steps:
           no-volume: true
           inherit-environment-vars: true
           vm-name: 10.14.6_6C_14G_80G
-          vm-registry-tag: "clean::cicd::git-ssh::nas::brew::buildkite-agent"
+          vm-registry-tag: "clean::cicd::git-ssh::nas::brew::buildkite-agent::${MACOS_10_14_TAG}"
```
kj4ezj, line 730: Nice work here! Good find, this is how we should have been doing it all along.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
+set -eo pipefail
+. ./.cicd/helpers/general.sh
+CDT_DIR_HOST=$(pwd)
```
kj4ezj, line 4: This `CDT_DIR_HOST` variable is a duplicate of the `ROOT_DIR` variable provided by `general.sh`. Let's use that instead.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+
+[[ ! -z "$CONTRACTS_VERSION" ]] || export CONTRACTS_VERSION="$(cat "$PIPELINE_CONFIG" | jq -r '.dependencies["eosio.contracts"]')"
+git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git 
+
+if [[ $(uname) == 'Darwin' ]]; then
+    export PATH=$CDT_DIR_HOST/build/bin:$PATH
+    cd build_eosio_contracts
+    cmake $CDT_DIR_HOST/eosio.contracts
+    make -j$JOBS
```
kj4ezj: We should print significant commands before running them so people know what the CI system is doing when they look at the logs without having to look at the source code. Something like this:
```bash
CMAKE="cmake '$CDT_DIR_HOST/eosio.contracts'"
echo "$ $CMAKE"
eval $CMAKE
MAKE="make -j $JOBS"
echo "$ $MAKE"
eval $MAKE
```

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+    # Load BUILDKITE Environment Variables for use in docker run
+    if [[ -f $BUILDKITE_ENV_FILE ]]; then
+        evars=""
+        while read -r var; do
+            evars="$evars --env ${var%%=*}"
+        done < "$BUILDKITE_ENV_FILE"
+    fi
+
+    eval docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS_EOSIO_CONTRACTS\"
```
kj4ezj: Same here. This command is very likely to be of interest to the users of our CI system, especially for debugging:
```bash
DOCKER_RUN="docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS_EOSIO_CONTRACTS\""
echo "$ $DOCKER_RUN"
eval $DOCKER_RUN
```

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+    cd $CDT_DIR_HOST/build/tests/unit/test_contracts
+    touch wasm_abi_size.json
+    PATH_WASM=$(pwd)
+    JSON=$(echo '{}' | jq -r '.')
+    echo '--- :arrow_up: Generating wasm_abi_size.log file'
+    for FILENAME in *.{abi,wasm}; do
+        FILESIZE=$(wc -c <"$FILENAME")
+        export value=$FILESIZE
+        export key="$FILENAME"
+        JSON=$(echo "$JSON" | jq -r "(.\"$key\") += (env.value | tonumber)")
```
kj4ezj: It is best to use `jq` to access environment variables instead of using BASH string construction, because `jq` queries often have a myriad of special characters. It is also consistent with how you access `value`:
```bash
JSON="$(echo "$JSON" | jq -r '.[env.key] += (env.value | tonumber)')"
```
This gets rid of the first pair of parenthesis and all the escape sequences, making it more readable. This also runs slightly faster.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+
+    cd $CDT_DIR_HOST/build_eosio_contracts/contracts
+    for dir in */; do
+        cd $dir
+        for FILENAME in *.{abi,wasm}; do
+            if [[ -f $FILENAME ]]; then
+                FILESIZE=$(wc -c <"$FILENAME")
+                export value=$FILESIZE
+                export key="$FILENAME"
+                JSON=$(echo "$JSON" | jq -r "(.\"$key\") += (env.value | tonumber)")
```
kj4ezj: Same here.
```bash
JSON="$(echo "$JSON" | jq -r '.[env.key] += (env.value | tonumber)')"
```

.cicd/eosio-contracts-build.sh
kj4ezj: Also, we should invoke `cmake` using a relative path.
```bash
CMAKE="cmake ../eosio.contracts"
echo "$ $CMAKE"
eval $CMAKE
```
We use a relative path when invoking `cmake` in all of our other build steps. I think it makes it easier to follow along with what the CI system is doing. It also allows developers to copy and paste commands into their terminal, since the absolute paths will not match our workstations.

.cicd/eosio-contracts-unittest.sh
```diff
@@ -0,0 +1,35 @@
...
+[[ ! -z "$CONTRACTS_VERSION" ]] || export CONTRACTS_VERSION="$(cat "$PIPELINE_CONFIG" | jq -r '.dependencies["eosio.contracts"]')"
+git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git 
+
+# Docker Commands
+ARGS=${ARGS:-"--rm --init -v $(pwd):$MOUNTED_DIR"}
+. $HELPERS_DIR/docker-hash.sh
+
+if [[ $BUILDKITE == true ]]; then
+    # Generate Base Images
+    $CICD_DIR/generate-base-images.sh
```
kj4ezj: We should not call the `generate-base-images.sh` script in the test steps.

This script creates docker base-images using the source code in the `.cicd/platforms` folder. However, here, this script will be a no-op because if these platforms files have changed, the build steps will already have built the new base-images by now and docker will pull them down for us.

.cicd/eosio-contracts-unittest.sh
```diff
@@ -0,0 +1,35 @@
...
+    # Load BUILDKITE Environment Variables for use in docker run
+    if [[ -f $BUILDKITE_ENV_FILE ]]; then
+        evars=""
+        while read -r var; do
+            evars="$evars --env ${var%%=*}"
+        done < "$BUILDKITE_ENV_FILE"
+    fi
+    
+    echo '--- :arrow_up: Sanity tests for eosio.contract wasm and abi files'
+    eval docker run $ARGS $evars $FULL_TAG bash -c \"$TEST_CONTRACTS_COMMAND\"
```
kj4ezj: Here, again, we should print significant commands before running them.
```bash
DOCKER_RUN="docker run $ARGS $evars $FULL_TAG bash -c \"$TEST_CONTRACTS_COMMAND\""
echo "$ $DOCKER_RUN"
eval $DOCKER_RUN
```
This way, the customer can see what the CI system is trying to do and replicate it locally for debugging their builds and tests.

.cicd/eosio-contracts-unittest.sh
```diff
@@ -0,0 +1,35 @@
...
+. ./.cicd/helpers/general.sh
+
+[[ ! -z "$CONTRACTS_VERSION" ]] || export CONTRACTS_VERSION="$(cat "$PIPELINE_CONFIG" | jq -r '.dependencies["eosio.contracts"]')"
+git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git 
+
+# Docker Commands
+ARGS=${ARGS:-"--rm --init -v $(pwd):$MOUNTED_DIR"}
+. $HELPERS_DIR/docker-hash.sh
+
+if [[ $BUILDKITE == true ]]; then
```
kj4ezj: I think this `if` statement should be removed. We only use this `[[ "$BUILDKITE" == 'true' ]]` guard to prevent `buildkite-agent` from being invoked when running the CI code on our local machines, but `buildkite-agent` is not being used in this script.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+    make -j$JOBS
+else #Linux
+    ARGS=${ARGS:-"--rm --init -v $(pwd):$MOUNTED_DIR"}
+    . $HELPERS_DIR/docker-hash.sh
+
+    PRE_CONTRACTS_COMMAND="export PATH=$MOUNTED_DIR/build/bin:$PATH && cd $MOUNTED_DIR/build_eosio_contracts"
+    BUILD_CONTRACTS_COMMAND="cmake $MOUNTED_DIR/eosio.contracts && make -j$JOBS"
+
+    # Docker Commands
+    if [[ $BUILDKITE == true ]]; then
```
kj4ezj: This guard can also be removed, as we are not invoking `buildkite-agent` here and it would make debugging easier if engineers can invoke these scripts locally.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,89 @@
...
+        while read -r var; do
+            evars="$evars --env ${var%%=*}"
+        done < "$BUILDKITE_ENV_FILE"
+    fi
+
+    eval docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS_EOSIO_CONTRACTS\"
+
+fi
+
+if [[ $BUILDKITE == true ]]; then
```
kj4ezj: This `if` statement can be moved, too. If you put it between line 78 and 79, then this metrics generation code can be tested locally.
```bash
echo $JSON >> wasm_abi_size.json
if [[ "$BUILDKITE" == 'true' ]]; then
    buildkite-agent artifact upload wasm_abi_size.json
```

pipeline.jsonc
```diff
@@ -1,10 +1,10 @@
...
-            "eosio": "develop"
-        }
+  "eosio-dot-cdt": {
+    "pipeline-branch": "master",
+    // dependencies to pull for cdt integration tests, by branch, tag, or commit hash
+    "dependencies": {
+      "eosio": "develop",
+      "eosio.contracts": "develop"
     }
+  }
```
kj4ezj: **Optional** - I don't like formatting changes like this because it obscures the line `git` history, which can explain why a value is being used when you do blames or use tools like [git lens](https://gitlens.amod.io). But, if you are going to change the whole file, we should remove the `pipeline-branch` key as it hasn't been used in years.

kj4ezj (changes requested):
Wow, this is fantastic work! Just a few notes below.

We should also test the GitHub Actions pipeline to make sure it works for our community contributors. Buildkite is used for internal contributors, while GitHub Actions is used for external contributors.

There are two ways you can test GitHub Actions:
1. Fork like you are an external contributor
   1. Fork this repo using your GitHub account
   1. Push an empty commit
   1. PR back to the CDT repo, and GitHub actions should kick off.
1. Temporary code edit
   1. Delete the conditional that prevents GitHub Actions from running for internal contributors on your branch
   1. Test GitHub Actions
   1. Restore the conditional on your branch

.cicd/eosio-contracts-build.sh
shahabi8: I use `CDT_DIR_HOST` to get the full path to `eosio.cdt` bin directory for MacOS. I need this to build `eosio.contracts`. So `ROOT_DIR` doesn't work here.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

shahabi8:
> Wow, this is fantastic work! Just a few notes below.
> 
> We should also test the GitHub Actions pipeline to make sure it works for our community contributors. Buildkite is used for internal contributors, while GitHub Actions is used for external contributors.
> 
> There are two ways you can test GitHub Actions:
> 
> 1. Fork like you are an external contributor
>    
>    1. Fork this repo using your GitHub account
>    2. Push an empty commit
>    3. PR back to the CDT repo, and GitHub actions should kick off.
> 2. Temporary code edit
>    
>    1. Delete the conditional that prevents GitHub Actions from running for internal contributors on your branch
>    2. Test GitHub Actions
>    3. Restore the conditional on your branch

I removed all the wasm related work from github workflow. Here is my PR link for testing github: https://github.com/EOSIO/eosio.cdt/pull/1194/checks

.cicd/eosio-contracts-unittest.sh
shahabi8: Done.

.cicd/eosio-contracts-unittest.sh
shahabi8: Done.

.cicd/eosio-contracts-unittest.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

pipeline.jsonc
shahabi8: I removed the `pipeline-branch`.

.cicd/eosio-contracts-build.sh
kj4ezj: I see what you mean. We will improve this in story [AUTO-1341](https://blockone.atlassian.net/browse/AUTO-1341).

.cicd/pipeline.yml
```diff
@@ -147,6 +147,150 @@ steps:
...
+  - label: ":aws: Amazon_Linux 2 - Toolchain Tests"
+    command:
+      - "buildkite-agent artifact download build.tar.gz . --step ':aws: Amazon_Linux 2 - Build' --agent-access-token $$BUILDKITE_AGENT_ACCESS_TOKEN && tar -xzf build.tar.gz"
+      - "./.cicd/toolchain-tests.sh"
+    env:
+      IMAGE_TAG: "amazonlinux-2"
+    agents:
+      queue: "automation-eks-eos-tester-fleet"
+    timeout: ${TIMEOUT:-10}
+    skip: ${SKIP_AMAZON_LINUX_2}${SKIP_LINUX}${SKIP_TOOLCHAIN_TESTS}
```
kj4ezj: The `Amazon_Linux 2 - Toolchain Tests` Buildkite job step is in here twice, [here](https://buildkite.com/EOSIO/eosio-dot-cdt/builds/4319#57f3597d-f8dc-4ec6-b5fc-547a7f6daa73) and [here](https://buildkite.com/EOSIO/eosio-dot-cdt/builds/4319#276c0943-0478-423a-acd2-a1bed1f0df1d). We should remove one of them, probably the first one since it is by itself.

.cicd/eosio-contracts-unittest.sh
```diff
@@ -0,0 +1,34 @@
...
+
+[[ ! -z "$CONTRACTS_VERSION" ]] || export CONTRACTS_VERSION="$(cat "$PIPELINE_CONFIG" | jq -r '.dependencies["eosio.contracts"]')"
+git clone -b "$CONTRACTS_VERSION" https://github.com/EOSIO/eosio.contracts.git 
+
+# Docker Commands
+ARGS=${ARGS:-"--rm --init -v $(pwd):$MOUNTED_DIR"}
+. $HELPERS_DIR/docker-hash.sh
+
+FULL_TAG='eosio/ci-contracts-builder:base-ubuntu-18.04-develop'
+TEST_CONTRACTS_COMMAND="cd $MOUNTED_DIR/build_eosio_contracts/tests && ctest -j $JOBS --output-on-failure -T Test"
```
kj4ezj: We should use the `echo`/`eval` pattern for the `ctest` commands, too, since our customers (your team) will want to see exactly which `ctest` command is being invoked while debugging tests.

I think this will work:
```bash
TEST_CONTRACTS_COMMAND="cd $MOUNTED_DIR/build_eosio_contracts/tests && CTEST='ctest -j $JOBS --output-on-failure -T Test' && echo \\\"$ \\\$CTEST\\\" && eval \\\$CTEST"
```
I tested it in my shell.

Screen Shot 2021-08-31 at 1 27 05 PM

I know this solution looks terrible and all the escape sequences are abhorrent to work with. In most EOSIO repos, we have created _simple_ user-facing build and test scripts in `/scripts` that the CI system can call inside the docker container or Anka image, eliminating the need to construct complex commands using BASH variables and then pass them into the container as a single string to `bash -c`. I checked [eosio.contracts](https://github.com/EOSIO/eosio.contracts) to see if we already had a simple test script we could invoke inside the container but, unfortunately, [we do not](https://github.com/EOSIO/eosio.contracts/tree/394c96a5b8f54e1976a6e6b96a1841b1dc08b52b/scripts). I have opened story [AUTO-1375](https://blockone.atlassian.net/browse/AUTO-1375) to update the CI patterns in that repo, including creating a `/scripts/test.sh` script that this pipeline could invoke. This would also ensure that the `cmake` commands being used to test contracts are always the same in both pipelines.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+JSON=$(echo '{}' | jq -r '.')
+echo '--- :arrow_up: Generating wasm_abi_size.json file'
+for FILENAME in *.{wasm,abi}; do
+    FILESIZE=$(wc -c <"$FILENAME")
+    export value=$FILESIZE
+    export key="$FILENAME"
+    JSON="$(echo "$JSON" | jq -r '.[env.key] += (env.value | tonumber)')"
+done
+
+cd ../../../../build_eosio_contracts/contracts
```
kj4ezj: **Optional:** You can use `pushd` and `popd` ([example](https://opensource.com/article/19/8/navigating-bash-shell-pushd-popd)) to eliminate the need for these unwieldy series of relative paths that must stay in lock-step with other `cd` commands.

So, line 54 would become...
```bash
pushd build/tests/unit/test_contracts
```
...and line 64 would no longer need the series of `..` because you can just call `popd` like this.
```bash
popd
```
There are a few other places it might be wise to use this pair of commands in this file, but this pair of `cd` commands really stands out as error-prone.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+    if [[ -f $BUILDKITE_ENV_FILE ]]; then
+        evars=""
+        while read -r var; do
+            evars="$evars --env ${var%%=*}"
+        done < "$BUILDKITE_ENV_FILE"
+    fi
+
+    DOCKER_RUN="docker run $ARGS $evars $FULL_TAG bash -c \"$COMMANDS_EOSIO_CONTRACTS\""
+    echo "$ $DOCKER_RUN"
+    eval $DOCKER_RUN
```
kj4ezj: Same here, as far as using the `echo`/`eval` syntax to show the customer what these two build steps are doing, if we can. This one will be harder than the last one because there is _another_ layer of BASH variable interpolation which will need _another_ layer of escape sequences....gross.

Instead of doing all that, let's just roll line 37...
```bash
COMMANDS_EOSIO_CONTRACTS="$PRE_CONTRACTS_COMMAND && $BUILD_CONTRACTS_COMMAND"
```
...into line 47:
```bash
DOCKER_RUN="docker run $ARGS $evars $FULL_TAG bash -c \"$PRE_CONTRACTS_COMMAND && $BUILD_CONTRACTS_COMMAND\""
```
This will eliminate one layer of variable interpolation, making it the same problem as the other peer review comment.

Now, we should be able to do this with line 26...
```bash
BUILD_CONTRACTS_COMMAND="CMAKE='cmake ../eosio.contracts' && echo \\\"$ \\\$CMAKE\\\" && eval \\\$CMAKE && MAKE='make -j $JOBS' && echo \\\"$ \\\$MAKE\\\" && eval \\\$MAKE"
```
...and this with line 34:
```bash
BUILD_CONTRACTS_COMMAND="CMAKE='cmake -DBUILD_TESTS=true $MOUNTED_DIR/eosio.contracts' && echo \\\"$ \\\$CMAKE\\\" && eval \\\$CMAKE && MAKE='make -j $JOBS' && echo \\\"$ \\\$MAKE\\\" && eval \\\$MAKE"
```
I think that should do it. Again, this is gross but we will fix it with contracts test scripts in story [AUTO-1375](https://blockone.atlassian.net/browse/AUTO-1375).

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+
+fi
+
+touch wasm_abi_size.json
+cd build/tests/unit/test_contracts
+JSON=$(echo '{}' | jq -r '.')
```
kj4ezj: **Optional:** The `-r` flag in `jq` here doesn't make sense. That flag removes quotes from string outputs, but we don't have strings here. You should remove it or replace it with `-c`, which means "compact" and prints the JSON with no whitespace. This is desirable when you are storing JSON in BASH variables because it prevents mistakes with BASH looking for inter-frame spacing in some contexts when you do or don't use quotes around variable names.

This unnecessary flag might be my fault, as I gave you some code samples that may have had it by accident.

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+echo '--- :arrow_up: Generating wasm_abi_size.json file'
+for FILENAME in *.{wasm,abi}; do
+    FILESIZE=$(wc -c <"$FILENAME")
+    export value=$FILESIZE
+    export key="$FILENAME"
+    JSON="$(echo "$JSON" | jq -r '.[env.key] += (env.value | tonumber)')"
```
kj4ezj: **Optional:** The `-r` flag on `jq` here should be a `-c` (see above comment).

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+
+cd ../../../../build_eosio_contracts/contracts
+for dir in */; do
+    cd $dir
+    for FILENAME in *.{wasm,abi}; do
+        if [[ -f $FILENAME ]]; then
+            FILESIZE=$(wc -c <"$FILENAME")
+            export value=$FILESIZE
+            export key="$FILENAME"
+            JSON="$(echo "$JSON" | jq -r '.[env.key] += (env.value | tonumber)')"
```
kj4ezj: **Optional:** The `-r` flag on `jq` here should be a `-c` (see above comment).

.cicd/eosio-contracts-build.sh
```diff
@@ -0,0 +1,92 @@
...
+        fi
+    done
+    cd ..
+done
+
+echo '--- :arrow_up: Uploading wasm_abi_size.json'
+cd ../../
+echo $JSON >> wasm_abi_size.json
```
kj4ezj: **Optional:** If you want the JSON to be expanded (with whitespace) in the artifact, you can do this to restore it:
```bash
echo "$JSON" | jq '.' >> wasm_abi_size.json
```
I would also recommend including "metrics" in the filename.
```bash
echo "$JSON" | jq '.' >> wasm-abi-size-metrics.json
```
These things are only necessary for us humans, and can be left out. Firefox will display JSON in a human-readable format by default. Chromium, VScode, and other editors will not.

kj4ezj (changes requested):
I am only kicking this back for the duplicate pipeline step (see below). Since I am kicking it back anyways, I found a few other things we should change if you have time. I look forward to hearing your thoughts.

.cicd/pipeline.yml
shahabi8: oops! thanks for catching this. Done.

.cicd/eosio-contracts-unittest.sh
shahabi8: Very helpful, done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

.cicd/eosio-contracts-build.sh
shahabi8: Done.

kj4ezj (approved):
🎖️

nksanthosh (approved):

PR #1179, opened 2021-07-21
Add package tests for Homebrew.

williamblevins:
## Change Description

Add package tests for Homebrew.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -104,7 +104,7 @@ steps:
     timeout: ${TIMEOUT:-120}
     agents:
       - "queue=mac-anka-large-node-fleet"
-    skip: $SKIP_MOJAVE
+    skip: ${SKIP_MACOS_10_14}
 
   - wait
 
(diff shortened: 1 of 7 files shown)


.cicd/pipeline.yml
```diff
@@ -104,7 +104,7 @@ steps:
     timeout: ${TIMEOUT:-120}
     agents:
       - "queue=mac-anka-large-node-fleet"
-    skip: $SKIP_MOJAVE
+    skip: ${SKIP_MACOS_10_14}
```
kj4ezj, line 107: I would rather change everything else from `SKIP_MACOS_10_14` to `SKIP_MOJAVE` than to move these to the long and not human-friendly names.

kj4ezj (approved):
🍎

.cicd/pipeline.yml
williamblevins: All the new branches (1.8.x and develop) are already the long ones, so this is actually consistent with the others and EOSIO.

PR #1196, opened 2021-08-28
Support std::array by adding ABI symbol T[N]

softprofe:
## Change Description

EPE 948 

The build crash was caused by std::array was not supported.  After this change, cdt build is OK. 
The running  of wasm action is partly support , std::array can be used inside action, in structure, or nested using in structure.
However using std::array as a parameter of action is not supported so far. 

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


tests/unit/test_contracts/array_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+   TABLE tests {
+      uint64_t                    id;
+      std::array<uint8_t,32>      str;
+      uint64_t primary_key()      const { return id; }
+   };
+   struct info {
+      int age;
+      std::string name;
+   };
+   typedef multi_index<name("tests"), tests> tests_table;
```
nksanthosh, line 22: Please add a KV variant of this test as well

tests/unit/test_contracts/array_tests.cpp
```diff
@@ -0,0 +1,125 @@
+#include <eosio/eosio.hpp>
```
nksanthosh, line 1: Please evaluate if this needs to be a toolchain test instead of a unit test (as this relates to compilation/abi gen failure)

tests/unit/test_contracts/array_tests.cpp
```diff
@@ -0,0 +1,125 @@
...
+      }
+   }
+ 
+   // test complex data
+   [[eosio::action]]
+   void testcom(name user) {
+      require_auth(user);
+      tests_table _tests(get_self(), get_self().value);
+
+      std::array<uint8_t, 32> str = {'a','a','a','a','a','a','a','a',
```
heifner, line 172: How about something like: `std::array<std::array<std::string, 5>, 13>`. Something more than just ints.

softprofe: Somethings like std::array<std::array<std::string, 5>, 13> is supported and works pretty good.

heifner: Please add a test to illustrates that.

tests/unit/test_contracts/array_tests.cpp
softprofe: After look at toolchain folder, I think add new tests under toolchain better than move this test to toolchain. now using_std_array.cpp and using_std_array.json were added under toolchain/build-pass, same as what std::tuple and std::variant did.

tests/unit/test_contracts/array_tests.cpp
softprofe: Added tests include using std::array<std::array<std::string, 4>, 2>

tests/unit/test_contracts/array_tests.cpp
bogniq: Is this test currently being used, or a placeholder to be used later in integration test?

softprofe: Some of theses actions are used and works well, as testin(for test array inside function), testcom( using array in multi index), testne(test nested inside function). however some of theses can't work for now as some function are not yet implemented(json to bin and bin to json, json to varient, varient to json,etc...), testepa( test array as parameter) testre(array as return) testkv(using array in kv table) as these action need data translation function so can't support for now. We will do with these in a new EPE, the test actions can't work now will be used in future.

tests/toolchain/build-pass/using_std_array.cpp
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#include <eosio/eosio.hpp>
+#include <eosio/print.hpp>
+#include <array>
+
+using std::array;
+using namespace eosio;
+
+class[[eosio::contract("hello")]] hello : public contract
```
bogniq: It's better to rename this contract to be the same as the file name (i.e. `using_std_array`), otherwise due to CDT change https://github.com/EOSIO/eosio.cdt/pull/1037, using `eosio-cpp` with this file as input and no options like `-o hello.wasm` will not generate the ABI file. 
Also, if this PR is more closely related with ABI generation, it can be put under `abigen-pass` so that the generated ABI can be checked.

tools/include/eosio/gen.hpp
```diff
@@ -541,6 +541,23 @@ struct generation_utils {
...
+      else if ( is_template_specialization( type, {"array"} )) {
+         std::string orig = type.getAsString();
+         std:: string ret = "";
+         ret += get_template_argument_as_string( type, 0 );
+         ret = replace_in_name(ret);
+         ret += '[';
+         auto pos1 = orig.find_last_of(',');
+         auto pos2 = orig.find_last_of('>');
+         std::string digits = orig.substr(pos1 + 1, pos2 - pos1 - 1);
+         for(auto & c : digits) {
```
bogniq: After getting substring `digits`, could it just strip (surround) spaces using some standard library function (and then append to `ret`)? If `digits` contains chars other than digits, some earlier phase will throw errors.

tests/toolchain/build-pass/using_std_array.cpp
softprofe: Sure, changed to using_std_array.

tools/include/eosio/gen.hpp
softprofe: Good idea,  std::remove is used instead of for loop.

tests/unit/test_contracts/array_tests.cpp
```diff
@@ -0,0 +1,192 @@
...
+   struct info {
+      int age;
+      std::string name;
+   };
+
+   typedef multi_index<name("tests"), tests> tests_table;
+   
+   struct my_struct {
+      uint32_t id;
+      std::array<std::array<std::string,4>,2> aastr;
```
plroblox: Better use typedef here for above array of arrays like typedef std::array<std::string,4> ary_str4 and then use array<ary_str4,2> as the type of above aastr. 
Because so far any abi-type ending with [][] would result in failure when tring to use cleos to put data into nodeos, e.g if using vector<vector<uint16_t> > vv, then using cleos to set vv as shown at https://github.com/EOSIO/nested-container/blob/main/nestcontn2a/nestcontn2a.cpp#L392 would result in  **Error 3015004: The type defined in the ABI is invalid, Error Details: uint16[][]**, the problem currently can be resolved by typedef vector<uint16_t> vec_uint16, and then use vector<vec_uint16> vv, the corrsponding type of vv would be vec_uint16[], instead of uint16[][]

bogniq (commented):
I would also suggest: 1. move the toolchain test `using_std_array` to `abigen-pass` and add json option `abi-file` pointing to the expected ABI file containing `"type": "int32[32]"` (see current `nested_container` under `abigen-pass` as an example); 2. shorten the PR title and move details into the PR description. Excellent work!

tests/unit/test_contracts/array_tests.cpp
softprofe: changed, use typedef std::array<std::string,4> array_string_4

softprofe:
> I would also suggest: 1. move the toolchain test `using_std_array` to `abigen-pass` and add json option `abi-file` pointing to the expected ABI file containing `"type": "int32[32]"` (see current `nested_container` under `abigen-pass` as an example); 2. shorten the PR title and move details into the PR description. Excellent work!

moved to abigen-pass

bogniq (approved):
After other related PRs in abieos are done, let's revisit the complete solution for std::array support.

PR #1197, opened 2021-08-30
Fix not working --no-missing-ricardian-clause

conr2d:
## Change Description

This PR makes `--no-missing-ricardian-clause` compile option work correctly.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tools/include/eosio/abigen.hpp
+++ b/tools/include/eosio/abigen.hpp
@@ -69,12 +69,8 @@ namespace eosio { namespace cdt {
          auto action_name = decl->getEosioActionAttr()->getName();
 
          if (!checked_actions.insert(get_action_name(decl)).second)
-            CDT_CHECK_WARN(!rcs[get_action_name(decl)].empty(), "abigen_warning", decl->getLocation(), "Action <"+get_action_name(decl)+"> does not have a ricardian contract");
-
-         if (!suppress_ricardian_warnings)
-            if (rcs[get_action_name(decl)].empty())
-               // TODO:
-               std::cout << "Warning, action <"+get_action_name(decl)+"> does not have a ricardian contract\n";
+            if (!suppress_ricardian_warnings)
+               CDT_CHECK_WARN(!rcs[get_action_name(decl)].empty(), "abigen_warning", decl->getLocation(), "Action <"+get_action_name(decl)+"> does not have a ricardian contract");
 
          ret.ricardian_contract = rcs[get_action_name(decl)];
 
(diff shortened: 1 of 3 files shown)


tools/include/eosio/abigen.hpp
```diff
@@ -729,6 +721,7 @@ namespace eosio { namespace cdt {
       public:
          explicit eosio_abigen_visitor(CompilerInstance *CI) {
             get_error_emitter().set_compiler_instance(CI);
+            set_suppress_ricardian_warning(ag.suppress_ricardian_warnings);
```
dimas1185: I don't think you need this one. `suppress_ricardian_warnings` above is taken from `abigen`

tools/include/eosio/abigen.hpp
```diff
@@ -96,12 +92,8 @@ namespace eosio { namespace cdt {
...
 
          if (!checked_actions.insert(get_action_name(decl)).second)
-            CDT_CHECK_WARN(!rcs[get_action_name(decl)].empty(), "abigen_warning", decl->getLocation(), "Action <"+get_action_name(decl)+"> does not have a ricardian contract");
-
-         if (!suppress_ricardian_warnings)
-            if (rcs[get_action_name(decl)].empty())
-               // TODO
-               std::cout << "Warning, action <"+get_action_name(decl)+"> does not have a ricardian contract\n";
+            if (!suppress_ricardian_warnings)
+               CDT_CHECK_WARN(!rcs[get_action_name(decl)].empty(), "abigen_warning", decl->getLocation(), "Action <"+get_action_name(decl)+"> does not have a ricardian contract");
```
dimas1185, line 96: what is the difference? Does it now outputs to stderr instead of stdout?

dimas1185:
thanks for your contribution.
do you mind to add toolchain tests, please?
you can take an example from here:
https://github.com/EOSIO/eosio.cdt/blob/4ff779217c54dccc8fb9a59463c0e1c6cb78e82b/tests/toolchain/compile-pass/warn_action_read_only.json

basically you need just to create dummy contract with same name and check stderr/stdout in json. you can use regexp there.
to run toolchain tests use `./build/tools/toolchain-tester/toolchain-tester`

tools/include/eosio/abigen.hpp
conr2d: Duplicated code. It seems original author intended replacing stdout warning message with CDT_CHECK_WARN macro when macro is implemented, but just added rather than replace it.

tools/include/eosio/abigen.hpp
conr2d: Yes, when I wrote this patch, `eosio_abigen_visitor` called `eosio_abigen_visitor::parse_contracts` instead of `abigen::parse_contracts`. This is obsolete now.

dimas1185: got it. Yes, After some refactoring ricardian contracts generation was broken and recently it was fixed including this warning suppression.
So now other then correcting warning message there is nothing to do. Anyway if you'll add toolchain tests for this warning it will be helpful for future

PR #1198, opened 2021-08-30
Add CICD pipeline documentation plus fix several MacOS-related CICD pipeline keys.

williamblevins:
## Change Description

Add CICD pipeline documentation plus fix several MacOS-related CICD pipeline keys.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline-upload.sh
+++ b/.cicd/pipeline-upload.sh
@@ -1,8 +1,24 @@
 #!/bin/bash
 set -eou pipefail
 
+# attach pipeline documentation
+echo '+++ :md: Attaching Documentation'
+export DOCS_URL="https://github.com/EOSIO/eosio.cdt/blob/${BUILDKITE_COMMIT:-master}/.cicd/README.md"
+export RETRY="$(buildkite-agent meta-data get pipeline-upload-retries --default '0')"
+buildkite-agent meta-data set pipeline-upload-retries "$(( $RETRY + 1 ))"
+if [[ "$BUILDKITE" == 'true' && "$RETRY" == '0' ]]; then
+    echo "This documentation is also available on [GitHub]($DOCS_URL)." | buildkite-agent annotate --append --style 'info' --context 'documentation'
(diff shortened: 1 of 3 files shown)


.cicd/pipeline-upload.sh
```diff
@@ -1,8 +1,24 @@
...
-envsubst "$VARS" < "./.cicd/pipeline.yml" > "./.cicd/pipeline.yml.out"
-buildkite-agent artifact upload ./.cicd/pipeline.yml.out
-buildkite-agent pipeline upload ./.cicd/pipeline.yml.out
+envsubst "$VARS" < ".cicd/pipeline.yml" > ".cicd/pipeline.yml.out"
+# pipeline upload
+echo '+++ :pipeline_upload: Deploying Pipeline Steps'
+if [[ "$BUILDKITE" == 'true' ]]; then
+    buildkite-agent artifact upload .cicd/pipeline.yml.out
+    buildkite-agent pipeline upload .cicd/pipeline.yml.out
+fi
```
kj4ezj: If we are going to use `envsubst` to explicitly call out the variables we are interpolating in the `pipeline.yml`, we should also pass `--no-interpolation` to the `buildkite-agent pipeline upload` step because it will fill in any variables we are not explicitly calling out in `envsubst`. If we are okay with `buildkite-agent` interpolating variables (I am apathetic, personally) then we should remove `envsubst` entirely. As-is, this is a misleading contradiction. It looks like the original code you are replacing was, too.

Separately, it would be great if the `pipeline.yml` we are attaching as a YAML file actually ends with the `*.yml` extension so that browsers, browser extensions, and editors are able to recognize it as YAML without having to rename the file or "tell" our tooling this is YAML.

kj4ezj (changes requested):
Should we fix this misleading pattern?

.cicd/pipeline-upload.sh
williamblevins: Yeah, its a double operation which is confusing.

kj4ezj (approved):
📝

PR #1207, opened 2021-09-15
Let wasm-ld support --no-abigen which is passed from eosio-cpp.

softprofe:
## Change Description

Let wasm-ld support --no-abigen which is passed from eosio-cpp
when using eosio-cpp --no-abigen xxxcontract.cpp, then wasm-ld pop error says -no-abigen is not unknown option, after this change, user can use eosio-cpp --no-abigen to compile xxxcontract.cpp without generating xxxcontract.abi, only get xxxcontract.wasm.

See also EPE 953

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/toolchain/build-pass/no_abigen_test.json
+++ b/tests/toolchain/build-pass/no_abigen_test.json
@@ -1,0 +1,10 @@
+{
+   "tests" : [
+     {
+       "compile_flags": ["--no-abigen"],
+       "expected" : {
+         "exit-code": 0,
+       }
+     }
+   ]
+}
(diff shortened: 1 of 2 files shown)


bogniq (commented):
A toolchain test can be added under "build-pass" for `--no-abigen` option (with a minimal test contract).

tests/toolchain/build-pass/no_abigen_test.json
```diff
@@ -0,0 +1,10 @@
+{
+   "tests" : [
+     {
+       "compile_flags": ["--no-abigen"],
+       "expected" : {
+         "exit-code": 0,
```
bogniq: Remove `,` will fix it.

softprofe: Yes, only one line inside {} should not has ','

bogniq (approved):

PR #1214, opened 2021-10-07
Add some privileged host functions wrappers

victorj8:
## Change Description

EOSIO 2.1 introduced a few new host functions to configure system wide parameters.

This change adds some host wrapper functions for some of them.

There are getter and setter for configuring WASM parameters:
- get_wasm_parameters_packed
- set_wasm_parameters_packed

And there are getter and setter for configuring the parameters of the new KV store feature:
- get_kv_parameters_packed
- set_kv_parameters_packed

Also for setting a limit resource on a specific account:
- get_resource_limit
- set_resource_limit

There is also the new getter and setter for configuring other blockchain parameters (as an alternative to get_blockchain_parameters_packed and set_blockchain_parameters_packed):

- get_parameters_packed
- set_parameters_packed

(Note: this is the only interface to access or modify the new max_action_return_value_size consensus parameter.)

For these last 2 host functions, as the new format for calling them is somewhat convoluted, it was decided to add the helper structure used on EOS: https://github.com/EOSIO/eos/blob/master/unittests/test-contracts/params_test/params_test.cpp, as a sort of wrapper to ease the use of the host functions.

Update (21/01/2022) : The get_parameters_packed and set_parameters packed host function wrappers were implemented on https://github.com/EOSIO/eos/pull/10974

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -185,6 +212,203 @@ namespace eosio {
...
+         char buffer[512];
+         datastream<char*> ds((char*)&buffer, sizeof(buffer));
+         ds << val;
+         ASSERT_LESS(ds.tellp(), sizeof(buffer));
+
+         packed += {buffer, ds.tellp()};
+         return *this;
+      }
+
+      void set() const{
```
bogniq: Is it possible to refactor struct `params_object` and add functions e.g. `set_parameters_packed(const eosio::params_object& params)` (and similar for the getter) outside of this struct?  (It's mentioned there is obstacle in doing so, please remind me what the difficulty is 🤔)

victorj8: I think it is possible, i copied this params_object directly from the eos example (linked in the description of the PR), do you think it would be better to put it on this other way?

bogniq: Since other similar functions are implemented on this other way (e.g. "set/get_wasm_parameters" uses struct `wasm_parameters`, and "set/get_blockchain_parameters" with `blockchain_parameters`), it seems to me more consistent in API design if users can call functions like "set/get_parameters_packed" with `params_object` as parameter. Please let me know your thoughts:)

victorj8: Yes, i agree to have this consistency in the way the API is called, i will make the changes.

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -5,6 +5,12 @@
 #include "../../core/eosio/name.hpp"
 #include "../../core/eosio/serialize.hpp"
 
+#define STR_I(x) #x
```
larryk85: These defines should live with check.hpp as it makes the most sense there.

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -18,6 +24,12 @@ namespace eosio {
          __attribute__((eosio_wasm_import))
          void set_resource_limits( uint64_t account, int64_t ram_bytes, int64_t net_weight, int64_t cpu_weight );
 
+         __attribute__((eosio_wasm_import))
+         int64_t get_resource_limit( uint64_t account, uint64_t resource );
```
larryk85, line 22: You should also have the C-API versions too.

libraries/eosiolib/contracts/eosio/privileged.hpp
larryk85: There is a nicer pattern than using macros that I came up with a while back.

Look at https://github.com/larryk85/cturtle/blob/main/include/bluegrass/cturtle/utils.hpp 

The call_info struct will capture the function, line and file info using the __builtin_LINE, _FUNCTION, and _FILE, etc. via default initialized parameters for it's constructor.

Then look at  https://github.com/larryk85/cturtle/blob/main/include/bluegrass/cturtle/log.hpp at one of the `format` functions to see how to use them.

You can simply create check with a call_info parameter and check_eq, check_less, and check_le.

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -185,6 +212,217 @@ namespace eosio {
...
+   };
+
+    /**
+     * Set the blockchain parameters (version using the set_parameters_packed host function)
+     *
+     * @ingroup privileged
+     *
+     * @param New blockchain parameters to set
+    */
+    inline void set_parameters(params_object& params) {
```
larryk85: This should be `const param_objects&`.  It's small but it can help the compiler to optimize a bit better as it doesn't know it the host function will mutate data or not.

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -185,6 +212,217 @@ namespace eosio {
...
+         (*this)(v);
+      }
+
+      params_object(const std::string& params) : packed(params) {}
+
+      bool operator == (const params_object& pp) const{
+         return packed == pp.packed;
+      }
+
+      params_object& operator () (unsigned_int val){
```
larryk85: What is the motivation to use the 'call operator' instead of the << operator?  It is a bit odd.

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -185,6 +212,217 @@ namespace eosio {
     */
    void get_blockchain_parameters(eosio::blockchain_parameters& params);
 
+   /**
+    *  Helper structure for host functions set_parameters_packed and get_parameters_packed
+    *  @ingroup privileged
+    */
+   struct params_object{
+      std::string packed;
```
larryk85: Why not use a 'vector' here?  The packing is a binary blob and not a guaranteed representable string?  Also because of strings small object optimization it throws some holes in RVO at times.

larryk85 (changes requested):

libraries/eosiolib/contracts/eosio/privileged.hpp
softprofe: Hi Victor,  I saw you packed  eosio::params_object as parameter of set_parameters_packed,  but however please follow the design document https://developers.eos.io/manuals/eos/v2.1/classeosio_1_1chain_1_1webassembly_1_1interface#function-get_blockchain_parameters_packed, that says: a packed data stream which represents an encoded sequence of parameter_id:paramter_value pairs with the following format: |varuint32:sequence_length | varuint32:parameter_id | :parameter_value | ... .   I know the eos part didn't strictly follow the design document, but this is a change to make it correct.

softprofe: And also the eosio::params_object is not efficient at all.  Using a signle datastream and 1 buffer to pack all data is efficient. but if using  each 1 data using a new buffer and new datastream, and then try move it to a string, that is not efficient.

And also  eosio::params_object  as parameter is hard to use by cleos input, did you ever think that?

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: I removed this code, as the get_parameters_packed and set_parameters packed host functions were implemented in a different way.

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: Added

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: I removed this code, as the get_parameters_packed and set_parameters packed host functions were implemented in a different way.

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: I removed this code, as the get_parameters_packed and set_parameters packed host functions were implemented in a different way.

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: I removed this code, as the get_parameters_packed and set_parameters packed host functions were implemented in a different way.

libraries/eosiolib/contracts/eosio/privileged.hpp
victorj8: I removed this code, as the get_parameters_packed and set_parameters packed host functions were implemented in a different way.

larryk85 (approved):

bogniq (approved):

PR #1217, opened 2021-10-08
Disable building with ccache for the libraries and tests directories

hmmblockone:
## Change Description

When ccache is installed, building CDT will fail. This PR addresses the issue by disabling ccache for the libraries and tests directory. This PR addresses cmake regression 1) in the issue: https://blockone.atlassian.net/browse/EPE-1161

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


bogniq (approved):
For now it's okay to disable ccache and sccache because they introduce more issues than benefits. A separate task can be created to investigate the issues if they need to be brought back later.

PR #1218, opened 2021-10-12
Move preactivate_feature() and is_feature_activated() to CDT, similar…

hmmblockone:
… to versions >= 1.7. Disable building with ccache for tests/ and libraries/ directories

## Change Description

This PR addresses two issues:

1) When ccache is installed, building CDT will fail. This PR addresses the issue by disabling ccache for the libraries and tests directory. This addresses cmake regression 1) in the issue: https://blockone.atlassian.net/browse/EPE-1161

2) Moves some code from later eosio.contracts to this version of CDT. This needed to accommodate changes to eosio.contracts 1.8.x in a separate PR, and brings CDT 1.6 in line with later version of CDT.  This is related to cmake regression 3) in the issue: https://blockone.atlassian.net/browse/EPE-1161

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


softprofe:
Checked with Santhosh, no requirement for 1.6.x, just close it.

PR #1232, opened 2021-11-06
Support std::array by adding ABI symbol T[N] - 1.8.x

softprofe:
## Change Description

This PR is for merge back  PR1196 to 1.8.x, it is for partly support std::array, std::array is allowed to be used inside function, but can't be used as parameters / return value/ kv table so far. 

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


plroblox (approved):
Looks good

PR #1238, opened 2021-11-12
Using mpair instead of pair for std::map element abi type

softprofe:
Using mpair instead of pair for std::map element abi type, so as to avoid conflicting with pair.
See EPE 1469 also.

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

After this PR,  the std::map  corresponding abi structure type is "mpair_A_B[]" instead of "pair_A_B[]".
The abi file made before this PR can still work with eos,  as this PR doesn't change eos.

### DIFF
--- a/tests/unit/test_contracts/explicit_nested_tests.cpp
+++ b/tests/unit/test_contracts/explicit_nested_tests.cpp
@@ -94,6 +94,20 @@ CONTRACT explicit_nested_tests : public contract {
    }
 
    [[eosio::action]]
+   // usage : cleos -v push action eosio vmistr '[[[{"key":12,"value":"map test"},{"key":34,"value":"passed"}]]]'  -p eosio@active
+   std::vector<std::map<int, std::string>>  vmistr(std::vector<std::map<int, std::string>> input){
+      std::vector<std::map<int, std::string>> output = input;
+      return output;
+   }
+
+   [[eosio::action]]
+   // usage : cleos -v push action eosio vpistr '[[{"first":12,"second":"pair test"},{"first":34,"second":"passed"}]]'  -p eosio@active
+   std::vector<std::pair<int, std::string>>  vpistr(std::vector<std::pair<int, std::string>> input){
+      std::vector<std::pair<int, std::string>> output = input;
+      return output;
+   }
+
+   [[eosio::action]]
    // usage : cleos -v push action eosio tup '[{"field_0":1,"field_1":2.0,"field_2":[4,5,6,7]}]'  -p eosio@active
    std::tuple<uint64_t, std::optional<float>, std::vector<int>>  tup(std::tuple<uint64_t, std::optional<float>, std::vector<int>>  input){
       std::tuple<uint64_t, std::optional<float>, std::vector<int>> output = input;
--- a/tests/unit/test_contracts/explicit_nested_tests.cpp
+++ b/tests/unit/test_contracts/explicit_nested_tests.cpp
@@ -108,6 +108,20 @@ CONTRACT explicit_nested_tests : public contract {
    }
 
    [[eosio::action]]
+   // usage : cleos -v push action eosio vmpiistr '[[{"key":12,"value":{"first":"map pair","second":"test passed"}}]]'  -p eosio@active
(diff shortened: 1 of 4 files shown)


tools/include/eosio/gen.hpp
```diff
@@ -534,7 +534,7 @@ struct generation_utils {
    }
 
    void translate_explicit_nested_map_or_pair(const clang::QualType& type, int depth, std::string & ret, const std::string & tname, bool & gottype){
-      ret += depth > 0 ? tname + "_" : "pair_";
+      ret += depth > 0 ? tname + "_" : tname == "map" ? "kvpair_" : "pair_";
```
plroblox: The purpose is to distinguish map and pair, I suggest change "kvpair_" to "map" when tname=="map", because the name 'kvpair_' is kind of misleading, people might think 'kvpair_' is only related to kv-table or kv-map, but obviously you are refering to a general std::map, which can be used in multi-index table

tests/unit/test_contracts/explicit_nested_tests.cpp
```diff
@@ -93,6 +93,20 @@ CONTRACT explicit_nested_tests : public contract {
...
+      return output;
+   }
+
+   [[eosio::action]]
+   // usage : cleos -v push action eosio vpistr '[[{"first":12,"second":"pair test"},{"first":34,"second":"passed"}]]'  -p eosio@active
+   std::vector<std::pair<int, std::string>>  vpistr(std::vector<std::pair<int, std::string>> input){
+      std::vector<std::pair<int, std::string>> output = input;
+      return output;
+   }
+
```
plroblox, line 123: I suggest you add at least two more cases here: map of pairs, pair of maps, see https://github.com/EOSIO/eosio.cdt/blob/develop/tests/toolchain/build-pass/nestcontn2a.cpp and search for set mp and setpm to know their input formats,  it is even better you can try to add map of maps, pair of pairs as shown in setmm and setpp of the same link

plroblox (commented):
Nice work! Has two suggestions

tests/unit/test_contracts/explicit_nested_tests.cpp
softprofe: Good idea.

tools/include/eosio/gen.hpp
softprofe: As map_A_B[] would looks more like vector<map>, after talk with Ping, we agree with using mpair_A_B to represent a element of map, so kvpair will change to mpair.

tests/unit/test_contracts/explicit_nested_tests.cpp
softprofe: Test cases added, and per my test it passed.

tools/include/eosio/abigen.hpp
```diff
@@ -434,10 +434,10 @@ namespace eosio { namespace cdt {
 
          if(inside_type_name[0] != "" && inside_type_name[1] != ""){
             ret += inside_type_name[0] + "_" + inside_type_name[1];
-            abidef.type = "pair_" + inside_type_name[0] + "_" + inside_type_name[1] + "[]";
+            abidef.type = "mpair_" + inside_type_name[0] + "_" + inside_type_name[1] + "[]";
 
             abi_struct kv;
-            kv.name = "pair_" + inside_type_name[0] + "_" + inside_type_name[1];
+            kv.name = "mpair_" + inside_type_name[0] + "_" + inside_type_name[1];
```
plroblox, line 440: Can above kv.name be the same as abidef.type with +"[]" being added at the end?

plroblox (commented):
See one comment

tools/include/eosio/abigen.hpp
softprofe: for kv.name this struct,  not [] in the end, it just mpair_A_B.   
More detail in abi:  mpair_A_B is a struct, it stand for an element of std::map, which has key and value as member.  and mpair_A_B[] is the type of std::map.

plroblox (approved):
Please do the merge until CICD build is complete

PR #1245, opened 2021-11-23
Using mpair instead of pair for std::map element abi type -1.8.x

softprofe:
Merge back PR 1238 to 1.8.x, using mpair instead of pair so as to solve conflict between map and pair. see also EPE 1764.
## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


plroblox (approved):
ok

PR #1254, opened 2021-12-16
Add user function get/set_parameters to support host function get/set_parameters_packed

softprofe:
See EPE665 also.
Add user function get/set_parameters to support host function get/set_parameters_packed, and also test contract for it, test actions passed.

For the eos part is hard to change as a released host function, so I changed cdt part format to adapt eos part now it works well without eos change.

note: before set contract, need active parameter feature firstly. just run "cleos system activate PREACTIVATE_FEATURE" , this feature is not activated by default.

cleos -v push action eosio get '[[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]]' -p eosio@active
executed transaction: 714d42fb2c5cf023b4d9c5cfcb153e2df9c64f5b1ad7a90fb72eff882ab5e63b  168 bytes  786 us
#         eosio <= eosio::get                   {"param_ids":[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]}
=>                                return value: [{"first":0,"second":["uint64",1048800]},{"first":1,"second":["uint32",1008]},{"first":2,"second":["...
>> id = 0, value= 1048800
>> id = 1, value= 1008
>> id = 2, value= 524290
>> id = 3, value= 12
>> id = 4, value= 500
>> id = 5, value= 20
>> id = 6, value= 100
>> id = 7, value= 200000
>> id = 8, value= 1000
>> id = 9, value= 150000
>> id = 10, value= 100
>> id = 11, value= 3600
>> id = 12, value= 600
>> id = 13, value= 3888000
>> id = 14, value= 524288
>> id = 15, value= 5
>> id = 16, value= 7
>> id = 17, value= 256

cleos -v push action eosio set '[[{"first":0,"second":["uint64",4448800]},{"first":1,"second":["uint32",1228]}, {"first":2,"second":["uint32",528890]},{"first":15,"second":["uint16",15]}, {"first":16,"second":["uint16",17]}]]' -p eosio@active
executed transaction: 4515d16856411413ae5c1c8b935328607fa50f674f3b5e54c050b56ce706f14a  1168 bytes  1000 us
#         eosio <= eosio::set                   {"params":[{"first":0,"second":["uint64",4448800]},{"first":1,"second":["uint32",1228]},{"first":2,"...
warning: transaction executed locally, but may not be confirmed by the network yet

cleos -v push action eosio get '[[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]]' -p eosio@active                                                                        executed transaction: a00680833af891a8155c45811a7e8038f6f1c83b72ff1a0ff3963ac0a799ff0d  1192 bytes  1000 us
#         eosio <= eosio::get                   {"param_ids":[0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17]}
=>                                return value: [{"first":0,"second":["uint64",4448800]},{"first":1,"second":["uint32",1228]},{"first":2,"second":["...
>> id = 0, value= 4448800
>> id = 1, value= 1228
>> id = 2, value= 528890
>> id = 3, value= 1024
>> id = 4, value= 700
>> id = 5, value= 50
>> id = 6, value= 120
>> id = 7, value= 250000
>> id = 8, value= 1500
>> id = 9, value= 230000
>> id = 10, value= 1000
>> id = 11, value= 1800
>> id = 12, value= 900
>> id = 13, value= 2592000
>> id = 14, value= 1048576
>> id = 15, value= 15
>> id = 16, value= 17
>> id = 17, value= 512
warning: transaction executed locally, but may not be confirmed by the network yet         ]

## API Changes
- [x] API Changes

## Documentation Additions
- [x] Documentation Additions

The ids from 0 to 17 have this meaning :
   enum {
      max_block_net_usage_id,
      target_block_net_usage_pct_id,
      max_transaction_net_usage_id,
      base_per_transaction_net_usage_id,
      net_usage_leeway_id,
      context_free_discount_net_usage_num_id,
      context_free_discount_net_usage_den_id,
      max_block_cpu_usage_id,
      target_block_cpu_usage_pct_id,
      max_transaction_cpu_usage_id,
      min_transaction_cpu_usage_id,
      max_transaction_lifetime_id,
      deferred_trx_expiration_window_id,
      max_transaction_delay_id,
      max_inline_action_size_id,
      max_inline_action_depth_id,
      max_authority_depth_id,
      max_action_return_value_size_id
   };

For the value of these ids,  max_block_net_usage is unit64_t, max_inline_action_depth and max_authority_depth is uint16_t, and others are all uint32_t.
The the data transform before/after pack/unpack to/from datastream:
size  uint32<->unsinged_int 
id     uint32<->unsinged_int
data  keep as uint16/32/64 
note: unsigned_int is a struct, not unsigned int

### DIFF
--- a/libraries/eosiolib/contracts/eosio/privileged.hpp
+++ b/libraries/eosiolib/contracts/eosio/privileged.hpp
@@ -28,6 +28,12 @@ namespace eosio {
          uint32_t get_blockchain_parameters_packed( char* data, uint32_t datalen );
 
          __attribute__((eosio_wasm_import))
+         void set_parameters_packed( char* data, uint32_t datalen );
+
+         __attribute__((eosio_wasm_import))
+         uint32_t get_parameters_packed( char* id, uint32_t idlen , char* data, uint32_t datalen  );
+
+         __attribute__((eosio_wasm_import))
          void set_kv_parameters_packed( const char* data, uint32_t datalen );
 
          __attribute((eosio_wasm_import))
@@ -169,6 +175,27 @@ namespace eosio {
       )
    };
 
+   enum {
+      max_block_net_usage_id,
+      target_block_net_usage_pct_id,
+      max_transaction_net_usage_id,
+      base_per_transaction_net_usage_id,
+      net_usage_leeway_id,
+      context_free_discount_net_usage_num_id,
+      context_free_discount_net_usage_den_id,
+      max_block_cpu_usage_id,
+      target_block_cpu_usage_pct_id,
+      max_transaction_cpu_usage_id,
+      min_transaction_cpu_usage_id,
+      max_transaction_lifetime_id,
+      deferred_trx_expiration_window_id,
+      max_transaction_delay_id,
+      max_inline_action_size_id,
+      max_inline_action_depth_id,
+      max_authority_depth_id,
+      max_action_return_value_size_id
+   };
+
    /**
     *  Set the blockchain parameters
     *
@@ -184,6 +211,23 @@ namespace eosio {
     *  @param params - It will be replaced with the retrieved blockchain params
     */
    void get_blockchain_parameters(eosio::blockchain_parameters& params);
+
+   /**
+    *  Set the blockchain parameters flexibly by id data pair vector
+    *
+    *  @ingroup privileged
+    *  @param params - New blockchain parameters to set, only id and data in the params will be set.
+    */
+   void set_parameters(const std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params);
+
+   /**
+    *  Retrieve the blolckchain parameters flexibly by ids
+    *
+    *  @ingroup privileged
+    *  @param param_ids - The id vecter in which ids are being queried from chain, see upper enum
+    *  @param params - It is output data with the retrieved blockchain params, before call it should be empty
+    */
+   void get_parameters(const std::vector<uint32_t> & param_ids,  std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params);
 
    /**
     *  Tunable KV configuration that can be changed via consensus
--- a/libraries/eosiolib/contracts/eosio/privileged.hpp
+++ b/libraries/eosiolib/contracts/eosio/privileged.hpp
@@ -196,6 +196,8 @@ namespace eosio {
       max_action_return_value_size_id
    };
 
+   using id_param_pairs_type = std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>;
+
    /**
     *  Set the blockchain parameters
     *
@@ -218,7 +220,7 @@ namespace eosio {
     *  @ingroup privileged
     *  @param params - New blockchain parameters to set, only id and data in the params will be set.
     */
-   void set_parameters(const std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params);
+   void set_parameters(const id_param_pairs_type & params);
 
    /**
     *  Retrieve the blolckchain parameters flexibly by ids
@@ -227,7 +229,7 @@ namespace eosio {
     *  @param param_ids - The id vecter in which ids are being queried from chain, see upper enum
     *  @param params - It is output data with the retrieved blockchain params, before call it should be empty
     */
-   void get_parameters(const std::vector<uint32_t> & param_ids,  std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params);
+   void get_parameters(const std::vector<uint32_t> & param_ids,  id_param_pairs_type & params);
 
    /**
     *  Tunable KV configuration that can be changed via consensus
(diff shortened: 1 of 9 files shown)


libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -169,6 +175,27 @@ namespace eosio {
...
+      min_transaction_cpu_usage_id,
+      max_transaction_lifetime_id,
+      deferred_trx_expiration_window_id,
+      max_transaction_delay_id,
+      max_inline_action_size_id,
+      max_inline_action_depth_id,
+      max_authority_depth_id,
+      max_action_return_value_size_id
+   };
+
```
plroblox, line 198: Can you put these enums or mention them in Document Addition of this PR so that people know what 0,1, to 17 mean?

libraries/eosiolib/eosiolib.cpp
```diff
@@ -65,6 +69,48 @@ namespace eosio {
       ds >> params;
    }
 
+   //  considering of runtime efficiency, using a estimated size to save the exact size calculating time
+   const int estimate_param_ids_buff_size = 128;
+   const int estimate_params_buff_size = 256;
```
plroblox, line 74: Is it possible to add EOS_ASSERT() to make sure ids_buff and buff always less than these consts?

plroblox (commented):
see my comments

libraries/eosiolib/contracts/eosio/privileged.hpp
softprofe: Sure.

libraries/eosiolib/eosiolib.cpp
softprofe: Good idea, I add eosio::check to check that, in this file otherwhere all use eosio::check.

plroblox (approved):
looks ok

libraries/eosiolib/contracts/eosio/privileged.hpp
```diff
@@ -185,6 +212,23 @@ namespace eosio {
     */
    void get_blockchain_parameters(eosio::blockchain_parameters& params);
 
+   /**
+    *  Set the blockchain parameters flexibly by id data pair vector
+    *
+    *  @ingroup privileged
+    *  @param params - New blockchain parameters to set, only id and data in the params will be set.
+    */
+   void set_parameters(const std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params);
```
bogniq: How about create an alias for `std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>` for here and other related code by `using`?

libraries/eosiolib/eosiolib.cpp
```diff
@@ -65,6 +69,75 @@ namespace eosio {
...
+   void set_parameters(const std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>& params) {
+      char buff[estimate_params_buff_size];
+      unsigned_int size = params.size();
+      if(params.size() == 0) return;
+      eosio::datastream<char *> ds( buff, sizeof(buff) );
+      ds << size;
+      for(const std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>> & p : params ){
```
bogniq: Maybe just use `const auto &`?

libraries/eosiolib/eosiolib.cpp
```diff
@@ -65,6 +69,75 @@ namespace eosio {
...
+      eosio::check( size <= sizeof(buff_params), "params buffer is too small" );
+      eosio::datastream<const char*> para_ds( buff_params, size );
+      unsigned_int para_size;
+      para_ds >> para_size;
+      eosio::check(id_size == para_size, "pairs returned doesn't match amount of id to query");
+      for(uint32_t i = 0; i < para_size; ++i){
+         unsigned_int _id;
+         std::variant<uint16_t, uint32_t, uint64_t>  var;
+         para_ds >> _id;
+         uint32_t id = _id;
```
bogniq, line 123: Could `_id` be just used below, instead of introducing `id`?

softprofe: unsigned_int is a struct, it is not unsigned int, I'm confused with this at first look also. so basically _id and id is different type. _id can pass value to id.  and also _id.value is uint32_t, but if using _id.value,  the code reader can't see the type by first look, so using uint32_t id = _id to give a clear idea to code reader what the type is.

libraries/eosiolib/contracts/eosio/privileged.hpp
softprofe: Sure.

libraries/eosiolib/eosiolib.cpp
softprofe: Ok.

libraries/eosiolib/eosiolib.cpp
bogniq: I saw `unsigned_int` has an overloaded `==` with `uint32_t` so `_id` might work, but yeah I agree the current code is more explicit.

libraries/eosiolib/eosiolib.cpp
```diff
@@ -65,6 +69,76 @@ namespace eosio {
       ds >> params;
    }
 
+   using params_type = std::vector<std::pair<uint32_t, std::variant<uint16_t, uint32_t, uint64_t>>>;
```
bogniq: Can we put `params_type` in "privileged.hpp" so it could be used at other places?

softprofe: It's Ok to put in "privileged.hpp", however, I think we need an different alias name to let users know the type is only for get/set_parameters. how do you think?

bogniq (approved):

PR #1259, opened 2021-12-30
CDT support Oracle Linux 8

softprofe:
## Change Description

CDT support oracle linux8 after eos support that ,  see EPE 1756 also.
## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/pipeline.yml
+++ b/.cicd/pipeline.yml
@@ -169,7 +169,7 @@ steps:
       - "queue=mac-anka-node-fleet"
     skip: ${SKIP_MACOS_10_15}${SKIP_MAC}${SKIP_WASM_METRICS}
 
-- label: ":oraclelinux: Oracle_Linux 8 - EOSIO.CONTRACTS Build"
+  - label: ":oraclelinux: Oracle_Linux 8 - EOSIO.CONTRACTS Build"
     command:
       - "buildkite-agent artifact download build.tar.gz . --step ':oraclelinux: Oracle_Linux 8 - Build' --agent-access-token $$BUILDKITE_AGENT_ACCESS_TOKEN && tar -xzf build.tar.gz"
       - "./.cicd/eosio-contracts-build.sh"
@@ -214,7 +214,7 @@ steps:
     timeout: ${TIMEOUT:-10}
     skip: ${SKIP_CENTOS_7}${SKIP_LINUX}${SKIP_UNIT_TESTS}
 
- - label: ":oraclelinux: Oracle Linux 8 - Unit Tests"
+  - label: ":oraclelinux: Oracle Linux 8 - Unit Tests"
     command:
       - "buildkite-agent artifact download build.tar.gz . --step ':oraclelinux: Oracle_Linux 8 - Build' --agent-access-token $$BUILDKITE_AGENT_ACCESS_TOKEN && tar -xzf build.tar.gz"
       - "./.cicd/tests.sh"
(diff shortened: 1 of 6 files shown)


.cicd/toolchain-tests.sh
```diff
@@ -15,11 +15,20 @@ else # Linux
...
     ARGS=${ARGS:-"--rm --init -v $(pwd):$MOUNTED_DIR"}
 
-    PRE_COMMANDS="cd $MOUNTED_DIR/build"
-    SET_PYTHON_PATH="export CICD_PYTHON_PATH=/usr/local/bin/python3"
-    TEST="./tools/toolchain-tester/toolchain-tester ../tests/toolchain/"
-
-    COMMANDS="$PRE_COMMANDS && $SET_PYTHON_PATH && $TEST"
+    if [[ "$IMAGE_TAG" == "oraclelinux-8" ]]; then
+        PRE_COMMANDS="cd $MOUNTED_DIR/build"
+        PACKAGE_COMMANDS="yum install -y python38"
```
bogniq: Is it possible to put `yum install -y python38` in the above "oraclelinux-8.dockderfile" (similar with eos repo), so that "oracle-linux-8" don't need to be specially handled here?

softprofe: > Is it possible to put `yum install -y python38` in the above "oraclelinux-8.dockderfile" (similar with eos repo), so that "oracle-linux-8" don't need to be specially handled here?

I'm afraid it doesn't work per my test.
See line 9 In docker file, the python38 and python3-devel are already installed.

I guess the toolchain-test machine is not the same as build machine.
Before I made change to toolchain-tests.sh, the pipeline failed, then I change toolchain-tests.sh just like centos8 doing and it works.

.cicd/pipeline.yml
```diff
@@ -158,6 +169,17 @@ steps:
...
+    command:
+      - "buildkite-agent artifact download build.tar.gz . --step ':oracle_linux: Oracle_Linux 8 - Build' --agent-access-token $$BUILDKITE_AGENT_ACCESS_TOKEN && tar -xzf build.tar.gz"
+      - "./.cicd/eosio-contracts-build.sh"
+    env:
+      IMAGE_TAG: "oraclelinux-8"
+    agents:
+      queue: "automation-eks-eos-tester-fleet"
+    timeout: ${TIMEOUT:-10}
+    skip: ${SKIP_ORACLE_LINUX_8}${SKIP_LINUX}${SKIP_WASM_METRICS}
+
```
williamblevins, line 182: This is missing a key definition. See the other `EOSIO.CONTRACTS Build` sections.

.cicd/toolchain-tests.sh
williamblevins: This is weird.
* https://github.com/EOSIO/auto-eks-buildfleet/blob/master/fleet/automation-eks-eos-tester-fleet.yaml#L69
* https://github.com/EOSIO/auto-eks-buildfleet/blob/master/docker/buildkite-eos-builder.Dockerfile

If you need python3 on the image, then just install python3? `.cicd/platforms/oraclelinux-8.dockerfile`....

These others already do ... https://github.com/EOSIO/eosio.cdt/blob/bf2a6134c3516dc51387587d5b5ffce84c0b2f2f/.cicd/platforms/amazonlinux-2.dockerfile#L6

williamblevins (changes requested):
I agree with the comment about python3 weirdness. I would avoid special casing the test logic if you can just add it to the image.

The rest looked fine to me.

.cicd/toolchain-tests.sh
softprofe: python38 and python3-devel were already in Oracle linux 8 dockerfile,  did you ever have a look into that?

softprofe: > Is it possible to put `yum install -y python38` in the above "oraclelinux-8.dockderfile" (similar with eos repo), so that "oracle-linux-8" don't need to be specially handled here?

As Qing's mention, add a soft link may point to the right place. ln -s blahblah

praphael: @softprofe  Python3 should already be in the image.  The problem appears to be that CICD_PYTHON_PATH is hardcoded to /usr/local/bin/python3.  But Oracle Linux 8 installs python3 in /usr/bin/python3.8.    A quick fix in the Dockerfile would be to soft link /usr/local/bin/python3 to /usr/bin//python3.8.  Then you don't have to modify .cicd/toolchain-tests.sh.  Personally I think CICD_PYTHON_PATH being hardcoded is a code smell, but a clean fix would require modifying several other Dockerfiles which would be out of the scope of this PR.

bogniq (approved):

williamblevins (approved):

PR #1262, opened 2022-01-19
Unify format of get/set_parameters_packed

softprofe:
## Change Description

See EPE 2001 also.

Currently there are 3 format of get/set_parameters_packed formats.
void set_parameters_packed( char* data, uint32_t datalen ) , 
void set_parameters_packed( const char* params, uint32_t params_size );
void set_parameters_packed(const char*, std::size_t);

unified to void set_parameters_packed( const char* , uint32_t );

For get_parameters_packed, unified to get_parameters_packed( const char* , uint32_t,  char* , uint32_t );

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/native/intrinsics.cpp
+++ b/libraries/native/intrinsics.cpp
@@ -34,10 +34,10 @@ extern "C" {
    void set_blockchain_parameters_packed( char* data, uint32_t datalen ) {
       return intrinsics::get().call<intrinsics::set_blockchain_parameters_packed>(data, datalen);
    }
-   uint32_t get_parameters_packed( char* id, uint32_t idlen, char* data, uint32_t datalen) {
+   uint32_t get_parameters_packed(const char* id, uint32_t idlen, char* data, uint32_t datalen) {
       return intrinsics::get().call<intrinsics::get_parameters_packed>(id, idlen, data, datalen);
    }
-   void set_parameters_packed( char* data, uint32_t datalen ) {
+   void set_parameters_packed(const char* data, uint32_t datalen ) {
       return intrinsics::get().call<intrinsics::set_parameters_packed>(data, datalen);
    }
    void set_kv_parameters_packed( const char* data, uint32_t datalen ) {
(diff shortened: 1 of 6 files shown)


tests/unit/test_contracts/parameter_tests.cpp
```diff
@@ -12,9 +12,9 @@ namespace eosio {
      __attribute__((eosio_wasm_import))
      uint32_t get_blockchain_parameters_packed(char*, uint32_t);
      __attribute__((eosio_wasm_import))
-     void set_parameters_packed(char*, uint32_t);
+     void set_parameters_packed(const char*, uint32_t);
```
bogniq: `set_parameters_packed` and (below) `get_parameters_packed` are not used in this test, are they needed here? ((I know this is not the problem this PR trying to fix, just want to check)

softprofe: The functions to be test are set/get_parameters, they will call set/get_parameters_packed, however,  include header  file   privileged.hpp should already declare there host functions. so this section will be removed to make the code more clear and tested OK to run.

libraries/native/intrinsics.cpp
```diff
@@ -34,10 +34,10 @@ extern "C" {
    void set_blockchain_parameters_packed( char* data, uint32_t datalen ) {
       return intrinsics::get().call<intrinsics::set_blockchain_parameters_packed>(data, datalen);
    }
-   uint32_t get_parameters_packed( char* id, uint32_t idlen, char* data, uint32_t datalen) {
+   uint32_t get_parameters_packed(const char* id, uint32_t idlen, char* data, uint32_t datalen) {
       return intrinsics::get().call<intrinsics::get_parameters_packed>(id, idlen, data, datalen);
    }
-   void set_parameters_packed( char* data, uint32_t datalen ) {
+   void set_parameters_packed(const char* data, uint32_t datalen ) {
       return intrinsics::get().call<intrinsics::set_parameters_packed>(data, datalen);
```
plroblox, line 41: so the prototype set_parameters_packed(const char* data, uint32_t datalen ) means the passed in 'data' will not be set or modified, is this correct?

plroblox (commented):
see one comment

libraries/native/intrinsics.cpp
softprofe: Correct.

softprofe: Correct.

bogniq (approved):
This fixes the inconsistency for the specified functions. A later PR can clean up other functions (e.g. `set_blockchain_parameters_packed`).

PR #1267, opened 2022-02-08
Modify input format for struct, map and pair in docs and tests

plroblox:
## Change Description

Make map,pair have conforming JSON inputs and use shortcut inputs for struct,map,pair, changes are in docs/ and toolchain/ only

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


bogniq (commented):
LGTM. I would suggest to modify the PR's title to be more concise (e.g. "Modify input format for struct, map and pair in docs and tests", just an idea), and then put details in the description.

plroblox:
Good idea, the title is updated, thanks!

bogniq (approved):

PR #1278, opened 2022-06-23
Fix #1277: Installed CDT uses toolchain from build directory

### LINKED ISSUE #1277: After installing CDT, WasmToolchain still uses toolchain in build directory
After installing the CDT, i.e. to `/opt/eosio-cdt`, the CMake modules in the installed path still contain references to the toolchain in the build root. For example, look in `/opt/eosio-cdt/lib/cmake/eosio.cdt/EosioWasmToolchain.cmake`, and you'll see it sets the compilers etc to the ones (probably no longer) in the build directory rather than the proper ones in `/opt/eosio-cdt/eosio.cdt/bin`. Contract builds fail unless the build directory is preserved.

nathanielhourt:
See https://github.com/EOSIO/eosio.cdt/issues/1277 for details.

## Change Description

CMake preparation of the installed CMake module configuration files interpolates the build directory into the configuration rather than the install directory. Fix this by properly setting the `CDT_ROOT_DIR` variable to the installed path before generating the configuration files.

## API Changes
None

## Documentation Additions
None

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,13 +57,11 @@ include(modules/ToolsExternalProject.txt)
 set(WASM_SDK_BUILD true)
 
 ### Configure the EosioWasmToolchain.cmakes
-set(CDT_ROOT_DIR ${CMAKE_BINARY_DIR})
+set(CDT_ROOT_DIR ${CDT_INSTALL_PREFIX})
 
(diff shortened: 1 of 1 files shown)


nathanielhourt:
Hmm, doesn't work in some environments.

